PR #16, opened 2019-02-02
Merge Dev

strengejacke:
I think we can merge this, or not?


strengejacke:
Ok, there seems to be a conflict with the get_data.R file. I try to resolve it.

PR #18, opened 2019-02-04
update

DominiqueMakowski:
- fixed the first paragraph of GLMMadaptative tests (mostly by adding flatten=TRUE, do you think there is a need for testing both list and vector versions?)
- renamed model_information.R to model_info.R to match the function


DominiqueMakowski:
This made me think of two things;

- in `model_info()`, for consistency purposes?
  - is_bin - > is_binomial
  - is_pois -> is_poisson
  - is_bayes -> is_bayesian
  - link_fun -> link_function
  - nobs -> n_obs
- Altough I agree with the functions returning lists, it feels a bit clunky in some cases. I am wondering whether we shouldn't keep a list by default for model_info, as this function is expected (unless specified) to return multiple elements, and a vector (flatten=TRUE by default) for specific demands (find_predictors, find_response, find_random). It would also better dissociate the use of the parent and the baby functions, otherwise people might end using only stuff like `model_info(m)$conditional`. And I reckon that when asking specifically for predictors, response and such, you expected a vector of the asked thing. What do you think?

strengejacke:
Yes, we can rename the return-values from `model_info()` to the longer, more clear wording.

> And I reckon that when asking specifically for predictors, response and such, you expected a vector of the asked thing. What do you think?

So you would suggest setting `flatten = TRUE` by default?

PR #22, opened 2019-02-06
add travis badges

DominiqueMakowski:
and update website


strengejacke:
I personally don't care about travis and would rather remove CI than tests... This seems to me a safer way to ensure package stability, or what would you say?

DominiqueMakowski:
I think that as long as the tests pass we can keep both all tests and CI...

In general, I agree that the most important thing is having tests run locally. However, having a passing CI makes the package "looks serious" (i.e., looks maintained) when someone finds it on github. Critically, codecov offers an easy look on which code lines are not tested. 

I guess we can try fixing the 3 failing tests, and if it is too much work or creates new problems, we re-deactivate travis. What do you think?

strengejacke:
Ok, sounds good to me. So what is Travis  complaining about the brms-tests?

strengejacke:
And what does codecov anyway?

DominiqueMakowski:
Let's start by merging this branch so we can have a test covered master as reference. Then we can uncomment the 3 tests and see what happens.

codecov shows all tested and non-tested lines (see [here](https://codecov.io/gh/easystats/insight/tree/master/R))

image

PR #25, opened 2019-02-09
find_/get_parameters

DominiqueMakowski:
- [ ] make it work for brms for bayestestR
- [ ] make it better


DominiqueMakowski:
I have missed your draft of the functions!

DominiqueMakowski:
I would rename find_parameter to find_parameterS (as it general there are more than one parameter, at least the intercept and a slope)?

DominiqueMakowski:
Naive question: why favour the term "conditional" over "fixed" or others?

strengejacke:
I was thinking of it makes sense to keep the naming convention from the other functions. But I wasn't sure, either, if it's better to call it fixed? However, the `fixef()` for glmmTMB for instance, returns named lists, with names conditional etc.

DominiqueMakowski:
I don't have a particularly strong opinion...  Although accurate mathematically, "conditional" always seemed kind of obscure and intuitive to me... But there are no real other options apparently (since fixed effects can refer to conditional parameters *and more*?)

strengejacke:
Yes, to zero-inflated parameters, and for brms also to simplex-parameters etc.

PR #33, opened 2019-02-11
nobs and link_fun renaming

DominiqueMakowski:


DominiqueMakowski:
Feel free to merge :)

strengejacke:
I'll close this w/o merging, because I have fixed some other stuff in `model_info.R` as well, which will likely cause conflicts when I commit my changes. But I have also included your changes, so nothing is lost... This is still due to the fact that I can commit changes for _insight_ only from my pc at home, while changes to _bayestestR_ or _paramters_ can only be done from my pc at work...

This is an annoying interference between GitHub and cloud synchronization... :-/

PR #37, opened 2019-02-12
Support htest

DominiqueMakowski:
I thought it would be more complete to include also the support of htests (which encompass correlations and t-tests). I drafted a first implementation, but I feel it still needs your magic hand (response (and predictor?) are not properly caught) 😅


strengejacke:
I would like to put this on hold until we have the first release on CRAN, and target this feature for insight 0.2. There are many objects that are of class htest, and we should take some care for implementation. Maybe we could even put this into an own branch?

DominiqueMakowski:
As you want, I am ok with delaying this.

DominiqueMakowski:
@strengejacke I didn't realise that I pushed the readme update on this branch

R/find_parameters.R
```diff
@@ -224,6 +227,19 @@ find_parameters.stanreg <- function(x, ...) {
   ))
 }
 
+#' @export
+find_parameters.htest <- function(x, ...) {
+  fe <- colnames(as.data.frame(x))
+
+  cond <- fe[grepl(pattern = "^(?!(b\\[|sigma|Sigma))", fe, perl = TRUE)]
+  rand <- fe[grepl(pattern = "^b\\[", fe, perl = TRUE)]
```
strengejacke: Do `htest`-objects have random effects, or even a `as.data.frame()`-method? This looks strange to me...

strengejacke:
As I modified some file this morning... Is the statement about successful passing checks and no conflicts up to date? Else, we might also include a preliminary draft for htest-objects as well.

strengejacke:
I'm still uncertain about use cases of `htest`-support. Can you give me a "real-life" example how _insight_-functions could support `htest`-objects, or where `htest`-functions would benefit from _insight_?

Probably parameters, but even this is difficult to extract.

See here: we have no information on parameter names:
``` r
str(t.test(extra ~ group, data = sleep))
#> List of 9
#>  $ statistic  : Named num -1.86
#>   ..- attr(*, "names")= chr "t"
#>  $ parameter  : Named num 17.8
#>   ..- attr(*, "names")= chr "df"
#>  $ p.value    : num 0.0794
#>  $ conf.int   : num [1:2] -3.365 0.205
#>   ..- attr(*, "conf.level")= num 0.95
#>  $ estimate   : Named num [1:2] 0.75 2.33
#>   ..- attr(*, "names")= chr [1:2] "mean in group 1" "mean in group 2"
#>  $ null.value : Named num 0
#>   ..- attr(*, "names")= chr "difference in means"
#>  $ alternative: chr "two.sided"
#>  $ method     : chr "Welch Two Sample t-test"
#>  $ data.name  : chr "extra by group"
#>  - attr(*, "class")= chr "htest"
```

Here, we only have a reference to variable names in a single character string, difficult to extract properly:

``` r
library(sjmisc)
data(efc)
str(t.test(efc$c12hour, efc$neg_c_7, data = sleep))
#> List of 9
#>  $ statistic  : Named num 18
#>   ..- attr(*, "names")= chr "t"
#>  $ parameter  : Named num 912
#>   ..- attr(*, "names")= chr "df"
#>  $ p.value    : num 2.98e-62
#>  $ conf.int   : num [1:2] 27.2 33.9
#>   ..- attr(*, "conf.level")= num 0.95
#>  $ estimate   : Named num [1:2] 42.4 11.8
#>   ..- attr(*, "names")= chr [1:2] "mean of x" "mean of y"
#>  $ null.value : Named num 0
#>   ..- attr(*, "names")= chr "difference in means"
#>  $ alternative: chr "two.sided"
#>  $ method     : chr "Welch Two Sample t-test"
#>  $ data.name  : chr "efc$c12hour and efc$neg_c_7"
#>  - attr(*, "class")= chr "htest"
```

DominiqueMakowski:
Well again it would be best to remove htest from this branch, as it might need some more discussion. I will try to 1. remove htest support, push again to this branch (and merge it?) and start a new branch for htest

DominiqueMakowski:
Let's take the discussion of full htest support in #37. If you don't mind, I just left the `model_info` support, as the is_correlation / is_ttest info would be quite useful in report 😏.

PR #45, opened 2019-02-16
Revise Readme

strengejacke:
Since the list of supported models is already rather long, we might switch to floating text instead of using a list? And I wrote out the function names, it's clearer,  I think.


DominiqueMakowski:
> Since the list of supported models is already rather long, we might switch to floating text instead of using a list? And I wrote out the function names, it's clearer, I think. 

Agree

PR #71, opened 2019-03-28
BayesFactor support

DominiqueMakowski:


DominiqueMakowski:
@strengejacke I am not sure how to properly add parameters and model_info related functions... I added most of the test cases (as BFBayesFactor class covers several different models) in the `test-BFBayesFactor.R` file

strengejacke:
I thought I already commented here from my mobile phone, but apparently it doesn't seem to be here...

strengejacke:
In short, what I wrote in a rather long posting (but am too lazy to repeat):
- I don't like Bayes Factors pretty much, but that's just a sidenote here ;-)
- I couldn't get your tests pass w/o errors
- I think implementing BayesFactor-class is as difficult as htest-class (from t.test() etc.), i.e. information is not directly available or easy to extract (like formulas etc.)
- we should plan this earliest for 0.3.0 is my suggestion

strengejacke:
Damn, I really like how beatiful the [website-help](https://easystats.github.io/insight/reference/get_variance.html) and [PDF](https://cran.r-project.org/web/packages/insight/insight.pdf) render for `?get_variance` :-D

DominiqueMakowski:
On a sidenote, it seems that `get_priors` has to be fixed when different columns are present:

e.g., `insight::get_priors(insight::download_model("stanreg_glm_1"))`

```
 Error in (function (..., deparse.level = 1)  : 
  number of columns of matrices must match (see arg 2) 
5.
(function (..., deparse.level = 1) 
.Internal(rbind(deparse.level, ...)))(prior_intercept = structure(c("normal", 
"0", "10"), .Dim = c(1L, 3L), .Dimnames = list(NULL, c("dist", 
"location", "scale"))), prior = structure(c("normal", "0", "2.5",  ... 
4.
do.call(rbind, l) at get_priors.R#37
3.
as.data.frame(do.call(rbind, l), stringsAsFactors = FALSE) at get_priors.R#37
2.
get_priors.stanreg(insight::download_model("stanreg_glm_1")) at get_priors.R#21
1.
insight::get_priors(insight::download_model("stanreg_glm_1"))
```

strengejacke:
That's an annoying bug, which shouldn't pass the checks, but unfortunately, this function was not covered by checks at all (except for the example)... :-/ I have added tests now, which I actually should have done before.

PR #80, opened 2019-04-22
Updates to the JOSS manuscript

pdwaggoner:
Here are my thoughts on a clean, succinct JOSS-style manuscript that introduces the package. Feel free to keep, revise, or discard anything here you see fit. Thanks!

### DIFF
--- a/paper.md
+++ b/paper.md
@@ -22,19 +22,61 @@ bibliography: paper.bib
 
 # Summary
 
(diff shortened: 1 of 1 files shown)


pdwaggoner:
Awesome! So glad you're happy with it. I will do that in just a second here. Its a great package; honored to put my name on it (in part). You and @DominiqueMakowski did some great work here!

PR #96, opened 2019-05-11
Update README.md

DominiqueMakowski:
small update just to add a paragraph describing how insight see things.

WIP


DominiqueMakowski:
@strengejacke Maybe we could reuse some of the vignette into the README

strengejacke:
and vice versa ;-)

DominiqueMakowski:
I've updated the README based on the vignette. I like this vignette a lot, I think it helps in understanding the components a lot. I was a bit surprised when I saw in the list random and slopes before parameters, terms and variables, but I guess it makes sense following the figures.

Maybe I would rename this vignette `insight` so it is accessible via "Get Started", which a new user would feel more confident clicking (rather than going to "Articles" and looking at some "Function Design" 👻 doc ^^).

DominiqueMakowski:
so you want me to merge :) ?

strengejacke:
So I simply rename the vignette to the package's name?

DominiqueMakowski:
yes

PR #125, opened 2019-08-09
adds more model objects to `is_model` function; modify tests

IndrajeetPatil:
# Description

This PR aims to expand on the regression models currently included in `is_model` function.

If you want to see executed examples from any of the newly included objects, they can be seen here in the context of my package:
https://indrajeetpatil.github.io/ggstatsplot/articles/web_only/ggcoefstats.html

# Proposed Changes

- Added more classes to `is_model` function
- Styled the code to have one model object per line which makes it easy to peruse them
- Fixed a test in `test-lm` because `nls` object is now considered a model object
- Get rid of a `warning` about the value initializations in `test-lm` file


strengejacke:
Or we add an argument, like `supported_only` or so.

strengejacke:
@DominiqueMakowski any thoughts? I think `is_model()` is already used by some other packages, or not? Must check...

IndrajeetPatil:
Ah, I see. 

Yeah, I would definitely vote for either renaming this function to `is_supported_model()` or adding two functions (`is_supported_model()` and `is_model()`). Depends on what you see is the utility of these functions **beyond** `insight`. 

I can definitely see myself using `is_model` in my own package to write a conditional statement:
```
if model
    do x
else
    do y
```

strengejacke:
Yes, that probably makes most sense

DominiqueMakowski:
a generic `is_model` indeed makes sense, we could then add a `is_model_supported` (so both show in autocompletion and to underline its subsidiary nature)

IndrajeetPatil:
Okay, cool.

I will add `is_model_supported` in this PR. The new `is_model_supported` will be identical to the old `is_model`. 

This is a breaking change though, so I will do my best to highlight this in `NEWS.md` and provide a rationale for the same. 

Any ideas on why builds are failing only on `R 3.5`?!

DominiqueMakowski:
> Any ideas on why builds are failing only on R 3.5?!

Because @strengejacke doesn't like travis 😁

DominiqueMakowski:
More seriously, I am not sure why, we have some general problems with testing on all platforms and all versions in most of the easyverse packages, for reasons that are beyond our understanding (sometimes it was caused by the requirement of rstan in brms/rstanarm that made it complicated)

strengejacke:
> Because @strengejacke doesn't like travis

Haha! I actually like it more than you, because I don't deactivate 80% of the Travis platforms. :-P

@IndrajeetPatil I guess there are sometimes problems on Travis' side, i.e. we get false negatives. Anyway, I usually activate more platforms to get more comprehensive results. If checks fail, you can quickly see if its due to Travis or your package. Looks like there are some issue with lme4-installation, so it's safe to ignore.

strengejacke:
Ok, we need to check the usage in these cases to see in how far this breaks existing code:

- [ ] https://github.com/easystats/bayestestR/blob/099d57479ccb6a53b75c82fd0d1271e3873db5ba/R/bayesfactor_models.R
- [ ] https://github.com/easystats/performance/blob/fc169d4fdfce9ae7095096f78c6710d9ab6a5b0a/R/compare_performance.R

IndrajeetPatil:
**TO DO**:

- [ ] Change docs for `is_model`
- [ ] Introduce `is_supported_model`
- [ ] Run `styler` on package
- [ ] Regenerate website docs for the changed functions
- [ ] Remove spelling errors and use `spelling` package for spellcheck

IndrajeetPatil:
@DominiqueMakowski, @strengejacke Lemme know if you need me to make any other changes.

IndrajeetPatil:
When `CI` checks are complete, do you mind merging it? I will add myself to `DESCRIPTION` in one of the future `PR`s.

IndrajeetPatil:
@strengejacke Is there something pending before this can be merged?

strengejacke:
In first instance, my time to review this PR is limted this week ;-) But I think I can merge it the next days. Then, could you rename `is_supported_model()` into `is_model_supported()`? I think you accidentally swapped it...

That would be all, I guess.

IndrajeetPatil:
Hah, no that was deliberate. I guess we are phrasing the question differently in our mind:

> Is this model supported? 

versus 

> Is this a supported model?

I will rename it.

strengejacke:
No, the reason is simply to have common prefixes to "bundle" functions in the alphabeticall ordered help-files, and for auto-completion-reasons:

https://github.com/easystats/insight/pull/125#issuecomment-520111649

But this is just our used workflow, feel free to comment on our reasoning behind function names. :-)

strengejacke:
> (I think you could add yourself in the DESCRIPTION)

strengejacke:
There _must_ be a reason you are ignoring @DominiqueMakowski comments ;-)

IndrajeetPatil:
- [ ] renamed function
- [ ] added myself to author's list
- [ ] updated website
- [ ] styled code

PR #134, opened 2019-09-08
find_statistic returns t-statistic for gam, nnest, and multinom objects

IndrajeetPatil:
Minor edit to `find_statistic` function to properly handle `gam`, `nnest`, and `multinom` objects

### DIFF
--- a/R/find_statistic.R
+++ b/R/find_statistic.R
@@ -42,6 +42,7 @@ find_statistic <- function(x, ...) {
       "drc",
       "feis",
       "felm",
+      "gam",
       "gamlss",
       "garch",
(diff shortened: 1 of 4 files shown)


IndrajeetPatil:
I know that `gam` can return both *t* and *F* statistic, but until `get_statistic` function is ready, this will be a temporary solution. This is better than returning a `NULL`.

strengejacke:
> I know that gam can return both t and F statistic

Yes, that's a bit annoying ;-) t for the "main coefficients", F for the smooth terms, if I recall right? No good idea how to best handle this.

PR #154, opened 2019-10-31
more models

IndrajeetPatil:


strengejacke:
Can you name some packages from those model classes you added? Maybe it's possible to add more support for these as well.

PR #157, opened 2019-11-15
style, spelling, roxygen

IndrajeetPatil:


strengejacke:
Why do you want to remove the words from the wordlist?

IndrajeetPatil:
Because `usethis::use_spell_check` can no longer find those words anywhere. 
I didn't do it manually; the function removed them.

strengejacke:
ah, got it! So they might be "listed" or ignored internally...

PR #214, opened 2020-08-16
more models; spell check

IndrajeetPatil:


tests/testthat/test-get_data.R
```diff
@@ -2,8 +2,9 @@ if (require("testthat") && require("insight") && require("lme4")) {
   data("cbpp")
   set.seed(123)
   cbpp$cont <- rnorm(nrow(cbpp))
-  m <- glmer(cbind(incidence, size - incidence) ~ poly(cont,2) + (1 | herd),
-             data = cbpp, family = binomial)
+  m <- glmer(cbind(incidence, size - incidence) ~ poly(cont, 2) + (1 | herd),
+    data = cbpp, family = binomial
+  )
```
strengejacke, line 8: That's a strange rule.

strengejacke: I suggest keeping the previous version.

IndrajeetPatil: Hmm, good catch. This might actually be a bug in `styler`.

IndrajeetPatil: Apparently, not. This is what the `tidyverse` style guide dictates:
https://github.com/tidyverse/style/issues/39

Do you still want me to revert this change?

strengejacke: In the end, it's just a test, but I wouldn't format my other code that way... And the rule you linked to is not "working" here, as `family` is also a named argument, but in the same line.

PR #247, opened 2020-11-11
minor change to follow broom convention for `htest`

IndrajeetPatil:
``` r
broom::tidy(t.test(extra ~ group, data = sleep))
#> # A tibble: 1 x 10
#>   estimate estimate1 estimate2 statistic p.value parameter conf.low conf.high
#>      <dbl>     <dbl>     <dbl>     <dbl>   <dbl>     <dbl>    <dbl>     <dbl>
#> 1    -1.58      0.75      2.33     -1.86  0.0794      17.8    -3.37     0.205
#> # ... with 2 more variables: method <chr>, alternative <chr>

heads <- rbinom(1, size = 100, prob = .5)
broom::tidy(prop.test(heads, 100))
#> # A tibble: 1 x 8
#>   estimate statistic p.value parameter conf.low conf.high method     alternative
#>      <dbl>     <dbl>   <dbl>     <int>    <dbl>     <dbl> <chr>      <chr>      
#> 1     0.59      2.89  0.0891         1    0.487     0.686 1-sample ~ two.sided
```

Currently, using `parameters` + `insight::standardize_names` leaves the `df` column as `df` instead of calling it `parameter` when `broom`-style is chosen.

This PR fixes that.

### DIFF
--- a/R/standardize_names.R
+++ b/R/standardize_names.R
@@ -104,6 +104,7 @@ standardize_names.parameters_distribution <- function(data, style = c("easystats
     cn[cn == "Effects"] <- "effect"
     cn[cn == "Response"] <- "response"
     cn[cn == "CI"] <- "ci.width"
+    cn[cn == "df"] <- "parameter"
     cn[cn == "df_error"] <- "df.error"
     cn[cn == "df_residual"] <- "df.residual"
     cn[cn == "n_Obs"] <- "n.obs"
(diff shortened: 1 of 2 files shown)


strengejacke:
I find it a bit odd to rename `df` to `parameter`, however, this is how the element is named in `t.test()` (though the `print()` prints `df`). Is there any other case where a "real" `df` column is named `df` in broom::tidy? There's a df-column in `glance()`...

IndrajeetPatil:
Yeah, on a second thought, this does seem suboptimal. We don't want `glance`'s `df` to be renamed to `parameter`. 

Plus, sometimes the `parameters` output will contain three columns: `parameter`, `parameter1`, `parameter2`, which becomes confusing. 

I think let's just retain this as `df`. We can tolerate this amount deviation from `broom` convention.

PR #261, opened 2020-12-10
lmtest::coeftest

vincentarelbundock:
this is paired with a commit to the `parameters` repo


strengejacke:
Adding a `get_parameters()` would be good, too, I think. (which would be something like `x[, 1]`, similar to `get_statistic()`.

vincentarelbundock:
Happy to add one if you prefer, but `get_parameters.default` seems to work already:

``` r
library(insight)
library(lmtest)

mod <- coeftest(lm(hp~mpg, mtcars))
get_parameters(mod)
#>     Parameter   Estimate
#> 1 (Intercept) 324.082314
#> 2         mpg  -8.829731
```

<sup>Created on 2020-12-10 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

strengejacke:
ah, ok, missed that.

strengejacke:
Looks good, you may add yourself as contributor to the DESCRIPTION file, if you like.

vincentarelbundock:
No worries. If I ever end up making more meaningful contributions in the future I'll add myself, but this is minor, so I think we can just go ahead with what's there.

Thanks for the quick review!

PR #281, opened 2021-01-04
add class attr

strengejacke:


strengejacke:
@DominiqueMakowski I don't see any failing tests here when adding the class-attr function...?

image

strengejacke:
@DominiqueMakowski just updated R and all packages, to make sure that everything works as expected, included the latest testthat features:

325488 image1

PR #293, opened 2021-02-03
changing convention; cf. https://github.com/tidymodels/broom/issues/984

IndrajeetPatil:
As I had suspected, `broom` doesn't actually have any convention about the CI column.

So I would prefer to default to `conf.level`, which is more in line with `tidymodels`'s argument naming.


strengejacke:
Yes, makes sense

PR #297, opened 2021-02-04
revert unnecessary change by `styler`

IndrajeetPatil:


IndrajeetPatil:
@lorenzwalthert Apologies for tagging you here, but I am having a hard time creating a reprex for this issue. 

Is it a bug in `styler` that it converts `\n` to `\\n` in the `roxygen` documentation? 
It is breaking our examples, and so we would like to know if there is a way to avoid this from happening.

Thanks.

lorenzwalthert:
`dontrun{}` is the problem. Reprex: Put that in a file and run styler on it:

```r
#' Data frame and Tables Pretty Formatting
#'
#' @examples
#' \dontrun{
#'   c("\nA yellow line", "yellow")
#' }
#' @export
NULL
```

IndrajeetPatil:
@lorenzwalthert You can raise the issue, because I don't think I can do a good job of describing the issue clearly enough. 

Thanks again for your feedback! :)

PR #301, opened 2021-02-13
add `format_statistical_approach` function

IndrajeetPatil:


strengejacke:
@DominiqueMakowski not 100% sure about the purpose, but could this also be a `format_*` function?

IndrajeetPatil:
Happy to rename this.

`format_statistical_method` actually sounds much better.

strengejacke:
I'm not sure if this function (name) is intuitive? What is a "statistical method"?

IndrajeetPatil:
Fair enough. Maybe `format_statistical_approach`? 

Since parametric, non-parametric, etc. are different statistical approaches?

DominiqueMakowski:
`statistical_approach` is a bit long indeed would be nicer with only one word / part.
`format_approach`? I don't see from the top of my head any other case in which we could need the word "approach" 🤔 and it's common to say the Bayesian approach etc. 
Another word could be `framework` but that's a bit too general? 

``

strengejacke:
I suggest waiting to merge this PR after we have submitted the curren _insight_ the next days, I think there's still some need for discussion. E.g., I still don't see the practical purpose. Why would a user expect a value of `"parametric"` when the input for the function is `"p"`? What would be a use-case?

IndrajeetPatil:
This is mostly useful for my R packages. I am not sure how generally useful this will be for others.

As the docs suggest:

```r
#' @description
#'
#' Relevant for packages (e.g., `ggstatsplot`, `statsExpressions`, etc.), where
#' different types of statistical methods are supported. This switch function
#' converts strings entered by users to a common pattern.
```

`"p"` is just one instance. Users can also misspell the arguments, e.g. `"parameteric"` instead of `"parametric"`.

This function covers all such contexts.

DominiqueMakowski:
It doesn't do any harm being merged as far as I can foresee (and we might find some need / ways to expand later too 🤷 )

strengejacke:
Please explain the rational:

``` r
insight::format_approach("p-value")
#> [1] "parametric"
insight::format_approach("kruskal")
#> [1] "parametric"
```

<sup>Created on 2021-02-14 by the [reprex package](https://reprex.tidyverse.org) (v1.0.0)</sup>

strengejacke:
This function more looks like a "spell correction" for a very limited set of cases.

IndrajeetPatil:
> This function more looks like a "spell correction" for a very limited set of cases.

It is, it has a very limited utility. And the documentation also makes it clear:

```r
#' @description
#'
#' Relevant for packages (e.g., `ggstatsplot`, `statsExpressions`, etc.), where
#' different types of statistical methods are supported. This switch function
#' converts strings entered by users to a common pattern.
#'
#' Possible outputs are:
#' \itemize{
#'  \item{parametric}
#'  \item{non-parametric}
#'  \item{robust}
#'  \item{bayesian}
#' }
```

It also makes it clear that it is mostly helpful for package developers. 
So I don't see any user running something like `insight::format_approach("p-value")`.

But happy to close this PR if it bothers you to have a function oriented primarily towards package developers, and not users.

strengejacke:
It's not about having a function that mainly focusses on developers, since this is insight's main purpose. But what the function actually does, is converting an arbitrary string input into 1 out of 4 possible strings, based on the initial letter. I'm not quite sure if this "helper function" justifies an exported, visible function.

IndrajeetPatil:
I thought its usage in two other packages that have over 15,000 monthly downloads justified it 😅 

But point taken. I will just retain them as helper functions in respective packages.

DominiqueMakowski:
@strengejacke is right when saying that the current implementing could potentially be handled by a 'match.arg()' or a 'switch()' function, that said I could see the potential of such a function enabled for models, similar to find_algorithm, which would return the "approach" used by the model (even though there are problematic edgecases that would exist but still, conceptually I see the relevance of such an idea). 
In that context, the current "spell correction" implementation could be the method for characters? But then it's true that we might want to avoid adding to much WIP functions to the API 🤷

PR #348, opened 2021-04-26
Deprecate `predict = "relation"` in favor of `predict = "expectation"`.

bwiernik:
Following up on recent discussion, I think this is clearer and it also more closely aligns with stan/Bayes terminology of linpred, epred, and predict. Deprecating `"relation"` for now gives users time to change syntax.

### DIFF
--- a/R/get_predicted.R
+++ b/R/get_predicted.R
@@ -6,17 +6,18 @@
 #'   second argument has to be a model).
 #' @param data An optional data frame in which to look for variables with which
 #'   to predict. If omitted, the data used to fit the model is used.
-#' @param predict Can be \code{"link"}, \code{"relation"} (default), or
+#' @param predict Can be \code{"link"}, \code{"expectation"} (default), or
 #'   \code{"prediction"}. This modulates the scale of the output as well as the
 #'   type of certainty interval. More specifically, \code{"link"} gives an
 #'   output on the link-scale (for logistic models, that means the log-odds
-#'   scale) with a confidence interval (CI). \code{"relation"} (default) also
+#'   scale) with a confidence interval (CI). \code{"expectation"} (default) also
 #'   returns confidence intervals, but this time the output is on the response
 #'   scale (for logistic models, that means probabilities). Finally,
 #'   \code{"predict"} also gives an output on the response scale, but this time
 #'   associated with a prediction interval (PI), which is larger than a
 #'   confidence interval (though it mostly make sense for linear models). Read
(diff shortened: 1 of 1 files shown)


DominiqueMakowski:
So in modelbased `estimate_relation()` would also become `estimate_expected()`?

One small thing that borderline bothers me is that we had `c("prediction", "relation", "link")` as possibilities, now `c("prediction", "expected", "link")`. First question is, should we allow for two spellings; either `"predicted"` or `"prediction"`? 

However, somehow these two versions are not totally equivalent to me. The former has the advantage of being more consistent with "expected" (finishing with "ed"). But on the other hand, I feel like "prediction" emphasizes more the fact that it's about prediction intervals etc., especially since the function is already named `get_predicted`. But if we restricted to `"prediction"`, should we go for `"expectation"` instead of "expected"? Or allow both? 😅

R/get_predicted.R
```diff
@@ -145,10 +146,11 @@ get_predicted.data.frame <- function(x, data = NULL, verbose = TRUE, ...) {
 #' @export
 get_predicted.lm <- function(x,
                              data = NULL,
-                             predict = "relation",
+                             predict = c("expectation", "link", "prediction", "relation"),
```
DominiqueMakowski, line 149: In general, I'm personally not a big fan of putting the whole vector of possibilities in the definition. It's a very weird R-thing and it makes it less clear what the actual default is (especially for people that have some experience with any other languages that look at the list and be like *??what??* - it's really not intuitive or transparent that the default is actually the first element of the list). Yes one could argue that "it gives all the possibilities at first glance and no need to look at the arg description", which is especially true in the context for experienced R users - which are likely to be the users of insight

So we can keep it as is, especially since @strengejacke uses this form quite often too I think

strengejacke: Yeah, but just because I didn't know about the `choices` argument in `match.arg()` ;-)

But indeed, if the list is too long, it might be better to mention the option in the description of the arguments. Not sure if it makes sense to have a consistent style here?

bwiernik:
I think we should do expectation and prediction only as the options here and the function names in modelbased. The intervals are uncertainty intervals on either the expectation or the observation-level predictions.

R/get_predicted.R
strengejacke: I think so, too. In most cases, this "list" of options is not too long.

DominiqueMakowski (approved):

strengejacke (approved):

PR #355, opened 2021-05-07
works?

IndrajeetPatil:


strengejacke:
I think you have to check this locally (with "correct" CRAN settings), or via win-builder. I don't think GitHub is properly configured to find issues here.

IndrajeetPatil:
Okay, I am sending this to winbuilder, and you should receive an email about it.

strengejacke:
Seems to work! Local checks and win-builder found no issues. Maybe we can submit with this "usecase"?

IndrajeetPatil:
Absolutely. If we don't get any further issues from CRAN or from users related to this, we can start using the function in other packages as well.

PR #356, opened 2021-05-07
standardize install checks

IndrajeetPatil:


R/compute_variances.R
```diff
@@ -143,9 +143,8 @@
 # calculate the variance components...
 #
 .get_variance_information <- function(x, faminfo, name_fun = "get_variances", verbose = TRUE, model_component = "conditional") {
-  if (!requireNamespace("lme4", quietly = TRUE)) {
-    stop("Package `lme4` needs to be installed to compute variances for mixed models.", call. = FALSE)
-  }
+  # installed?
+  check_if_installed("lme4")
```
strengejacke, line 147: These are situations where probably we might use a specific, and no standardized message (so users know why there is an error). What would you say if we have an argument with an optional own message text?

IndrajeetPatil: Sounds like a good idea.

So, here
 https://github.com/easystats/insight/blob/983821069009044cbe9151813ac3d5c54d7c26ae/R/check_if_installed.R#L16

we can break the phrase with the second component being `reason = "for this function to work"` as a default?

```r
 "Package '", package, "' is required", reason, ".\n"", 
```

Issue #375, opened 2021-05-29
testing `model_info()` for rstanarm for different model families


strengejacke:
I fixed an issue with `model_info()` and `stanreg`, which then broke `model_info()` and `stan_polr()` (which now is on CRAN :-/

So we should check different models and families and see if they work with model_info. Everyone is invited to check one or two functions :-)

- [ ] stan_glm - family xy
- [ ] stan_aov
- [ ] stan_polr
- [ ] stan_betareg
- [ ] stan_biglm
- [ ] stan_clogit
- [ ] stan_gamm4
- [ ] stan_glm.nb
- [ ] stan_glmer
- [ ] stan_glmer.nb
- [ ] stan_lmer
- [ ] stan_lm
- [ ] stan_jm
- [ ] stan_nlmer
- [ ] stan_mvmer

IndrajeetPatil:
Since Daniel has removed tests I had written in https://github.com/easystats/insight/commit/d06cbcc1440c53cbd9a6d962680d91182124d017, we are back to square 1 now on this issue.

strengejacke:
Yes, I still don't like snapshot tests and am not convinced that these reliably detect bugs.

bwiernik:
Could you elaborate on that? The snaps just generate the (long) lists of model information. @IndrajeetPatil did you check that the information in each generated list was correct?

IndrajeetPatil:
> @IndrajeetPatil did you check that the information in each generated list was correct?

To the best of my ability, yes.

But this issue is not just about these particular tests, but more broadly about snapshot tests. The snapshot tests in `parameters` were also removed:
https://github.com/easystats/parameters/commit/fcf31a2836cc9bdc20383affeac91eb43faf86d7

I personally think these tests are actually much stricter than other manual tests we write because we are usually checking only certain parts of the output, while snapshots capture the entire output. So any minute change will also be detected.

Btw, since `vdiffr 1.0`, visual regression tests (being used in `see` package) also use the same strategy.

Relying on them _exclusively_ might be a bad idea, I agree, because they are by default not run on CRAN. But in contexts where generating expected conditions is going to be labor-intensive, they are a lifesaver.

strengejacke:
> Could you elaborate on that?

[...]

PR #378, opened 2021-05-29
adds tests for #375

IndrajeetPatil:
and style

### DIFF
--- a/tests/testthat/test-rstanarm.R
+++ b/tests/testthat/test-rstanarm.R
@@ -1,8 +1,8 @@
 .runThisTest <- Sys.getenv("RunAllinsightTests") == "yes"
-
+.runThisTest <- TRUE
 if (.runThisTest) {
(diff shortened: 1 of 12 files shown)


tests/testthat/test-rstanarm.R
```diff
@@ -1,8 +1,8 @@
 .runThisTest <- Sys.getenv("RunAllinsightTests") == "yes"
-
+.runThisTest <- TRUE
```
strengejacke, line 2: why did you set this to TRUE?

IndrajeetPatil: Oops!

I had set it temporarily for the tests to be run on my laptop but forgot to remove it.

Otherwise these tests are just skipped on my Mac.

IndrajeetPatil: I have reverted this change in `master` already.

PR #386, opened 2021-06-08
Make digits="apa" compliant with APA 7th edition

bwiernik:
APA now strongly recommends/requires exact p values unless p < .001.

### DIFF
--- a/R/format_p.R
+++ b/R/format_p.R
@@ -47,7 +47,9 @@ format_p <- function(p, stars = FALSE, stars_only = FALSE, name = "p", missing =
 
   if (digits == "apa") {
     digits <- 3
-  } else if (is.character(digits) && grepl("^scientific", digits)) {
+  }
+  
+  if (is.character(digits) && grepl("^scientific", digits)) {
     digits <- tryCatch(
       {
         as.numeric(gsub("scientific", "", digits, fixed = TRUE))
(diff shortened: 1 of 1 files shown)


R/format_p.R
```diff
@@ -46,17 +46,7 @@ format_p <- function(p, stars = FALSE, stars_only = FALSE, name = "p", missing =
...
-            ifelse(p > .999, "> .999",
-              paste0("= ", format_value(p, 3))
-            )
-          )
-        )
-      )
-    )
+    digits <- 3
   } else if (is.character(digits) && grepl("^scientific", digits)) {
     digits <- tryCatch(
```
strengejacke, line 51: the `if` statement previously returned `text`, the formatted p-value. I think the revision you're suggesting work when you don't put the first if-statement into an if-else-chain, but just start in line 50 with 

`if (is.character(digits) && grepl("^scientific", digits)) {`

strengejacke (changes requested):

R/format_p.R
strengejacke: and simply `if (digits == "apa") digits <- 3`.

bwiernik: ah thanks, yeah, misread something there

strengejacke:
looks good, thanks!

Issue #400, opened 2021-07-14
cleanup for afex v1.0.0


mattansb:
``` r
library("afex")

data(obk.long, package = "afex")

obk.long$treatment <- as.character(obk.long$treatment)
obk.long$phase <- as.character(obk.long$phase)

Mc <- afex::aov_car(value ~ treatment * gender + age + Error(id / (phase * hour)),
                    factorize = FALSE,
                    data = obk.long, include_aov = FALSE)
#> Contrasts set to contr.sum for the following variables: treatment, gender

M <- afex::aov_car(value ~ treatment * gender + Error(id / (phase * hour)),
                   data = obk.long, include_aov = FALSE)
#> Converting to factor: treatment
#> Contrasts set to contr.sum for the following variables: treatment, gender

B <- afex::aov_car(value ~ treatment * gender + Error(id),
                   data = obk.long, include_aov = FALSE)
#> Converting to factor: treatment
#> Warning: More than one observation per cell, aggregating the data using mean
#> (i.e, fun_aggregate = mean)!
#> Contrasts set to contr.sum for the following variables: treatment, gender

W <- afex::aov_car(value ~ Error(id / (phase * hour)),
                   data = obk.long, include_aov = FALSE)

mods <- list(Mc, M, B, W)
```
All of these work:

```R
# sapply(mods, find_algorithm)
# sapply(mods, find_formula)
# sapply(mods, find_interactions)
# sapply(mods, find_offset)
# sapply(mods, find_random_slopes)
# sapply(mods, find_smooth)
# sapply(mods, find_variables, simplify = FALSE)
# 
# sapply(mods, get_auxiliary) # why warning?
# sapply(mods, get_data)
# sapply(mods, get_df)
# sapply(mods, get_intercept)
# sapply(mods, get_loglikelihood)
# sapply(mods, get_modelmatrix)
# sapply(mods, get_predicted)
# sapply(mods, get_residuals)
# sapply(mods, get_sigma)
# sapply(mods, get_variance)
# sapply(mods, get_varcov)
# sapply(mods, get_deviance)
# sapply(mods, get_family, simplify = FALSE)
# 
# sapply(mods, find_parameters)
# sapply(mods, get_parameters)
# sapply(mods, find_predictors)
# sapply(mods, get_predictors)
# sapply(mods, find_random)
# sapply(mods, get_random)
# sapply(mods, find_response)
# sapply(mods, get_response)
# sapply(mods, find_statistic)
# sapply(mods, find_weights)
[...]

PR #401, opened 2021-07-14
fixes to insight when include_aov = F

mattansb:
Waiting for #400


strengejacke:
@mattansb We have now

```
── Failure (test-aovlist.R:72:5): find_terms ───────────────────────────────────
find_terms(m1) (`actual`) not equal to list(response = "yield", conditional = c("N", "P", "K", "Error(block)")) (`expected`).

`actual` is length 3
`expected` is length 2

`names(actual)`:   "response" "conditional" "error"
`names(expected)`: "response" "conditional"        

`actual$conditional`:   "N" "P" "K" "N:P"          "N:K" "P:K" "N:P:K"
`expected$conditional`: "N" "P" "K" "Error(block)"                    

`actual$error` is a character vector ('Error(block)')
`expected$error` is absent
── Failure (test-aovlist.R:76:5): find_terms ───────────────────────────────────
find_terms(m1, flatten = TRUE) (`actual`) not equal to c("yield", "N", "P", "K", "Error(block)") (`expected`).

`names(actual)` is a character vector ('response', 'conditional1', 'conditional2', 'conditional3', 'conditional4', ...)
`names(expected)` is absent

`actual[2:9]`:   "N" "P" "K" "N:P" "N:K" "P:K" "N:P:K" "Error(block)"
`expected[2:5]`: "N" "P" "K"                           "Error(block)"
```

(see https://github.com/easystats/insight/pull/401/checks?check_run_id=3083533468) - is this expected?

mattansb:
Right, I made a change to `find_terms` for `afex_aov` and `aovlist` objects to now return the error term under `$error` (it was previously giving weird results for complex error terms).

Why does this :

``` r
insight::find_terms(y ~ a * b)
#> $response
#> [1] "y"
#> 
#> $conditional
#> [1] "a" "b"
```

<sup>Created on 2021-07-18 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

Not also give `a:b` as a conditional term?

This is what `terms` gives:

``` r
attr(terms(y ~ a * b), "term.labels")
#> [1] "a"   "b"   "a:b"
```

<sup>Created on 2021-07-18 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

strengejacke:
Can we merge this?

mattansb:
No, still have my question about terms https://github.com/easystats/insight/pull/401#issuecomment-881990248

strengejacke:
I think `find_terms()` was always intended to return all predictors/variables, while `find_parameters()` will return the parameters.

we have 
- `find_variables()`, which returns the "pure" variable names
- `find_terms()`, which returns the variable names, including patterns like `log()` etc.
- `find_parameters()`, which should return the parameters names similar to what `coef()` / `summary()` gives
-

mattansb:
So it is intended that `find_terms` ignores interaction terms?

mattansb:
@strengejacke I'm done (:

strengejacke:
> So it is intended that find_terms ignores interaction terms?

Yes, the main intention was to find all "single" variables, but in that way how they entered the model (`log`, `I`, `scale`, ...).

Issue #420, opened 2021-09-12
Warning when `get_predicted`'s `predict` argument is unknown


vincentarelbundock:
Hi,

I've been using `insight` extensively in the past few weeks, and I have to say: this is a *killer* package. Truly a life save. I'm sooo grateful that you all put in the work to build this infrastructure. Awesome!

I was wondering if it wouldn't be a good idea to implement a "safer" default behavior for models that haven't been given their own `get_predicted.CLASSNAME` methods. For example:

``` r
library("insight")
library("pscl")
data("bioChemists", package = "pscl")
mod <- hurdle(art ~ phd + fem | ment, data = bioChemists, dist = "negbin")

# reponse gives the same results:
x <- insight::get_predicted(mod)
y <- predict(mod, type = "response")
all(x == y)
#> [1] TRUE

# zero does different results without warning
x <- insight::get_predicted(mod, predict = "zero", verbose = TRUE)
y <- predict(mod, type = "zero")
all(x == y)
#> [1] FALSE
```

Another downside of the current implementation is that many useful prediction types are currently unavailable, because they haven't been given explicitly in `insight`. In the example above, the `zero` predictions give us a useful quantity, and it's a shame that `response` is returned by default like this.

Thoughts?

vincentarelbundock:
I just realized that `get_predicted` default passes the `...` ellipses, so I could in principle just use `get_predicted(type="zero")`. But that's not ideal because it requires users to know which models go through the default method and which are actually supported.

One option would be to be less strict about the inputs of `.get_predicted_args`, and to call that from the default method.

I just put together a WIP pull request just to illustrate the point: https://github.com/easystats/insight/pull/421

(It does not work yet because `get_predicted.default` does not use `.get_predicted.args`)

bwiernik:
Thanks for raising the issue!

In terms of what is happening here, the `pscl:::predict.hurdle()` method uses a `type` argument to specify the kind of predictions, not a `predict` argument. When `predict.default()` passes the model and arguments to `stats::predict()` and it is passed on to the "hurdle" method, the `predict` argument is ignored 

The straightforward solution is that we should pass the contents of `predict` to `type` in `predict.default()`, as that is the most common name for this argument. We could do something more elaborate, such as checking to be sure the predict method has a `type` argument, but that starts to get very complex with essentially recreating the S3 dispatch system. I think just setting `type` will cover 99%+ of possible cases.

PR #422, opened 2021-09-12
Set `type` argument in `get_predicted.default()`

bwiernik:
Closes https://github.com/easystats/insight/issues/420
Closes https://github.com/easystats/insight/pull/421


vincentarelbundock:
Oh yeah, this is *much* better than what I suggested.


Solid work!

Issue #423, opened 2021-09-13
`get_predicted` error when `newdata` does not include all factor levels


vincentarelbundock:
In this model, `cyl` is a factor variable with 3 levels. I create a `nd` dataset to use for prediction, but this dataset only includes 2 rows -- thus excluding at least one of the factor levels from the prediction data. This dataset yields correct predictions with the standard `predict` function, but not with `get_predicted`:

``` r
library(insight)

mod <- lm(mpg ~ factor(cyl), data = mtcars)

nd <- structure(list(cyl = structure(2:3, .Label = c("4", "6", "8"), class = "factor")), out.attrs = list( dim = c(cyl = 2L), dimnames = list(cyl = c("cyl=6", "cyl=8"))), class = "data.frame", row.names = c(NA, -2L))

nd
#>   cyl
#> 1   6
#> 2   8

predict(mod, newdata = nd, se.fit = TRUE, interval = "confidence")
#> $fit
#>        fit      lwr      upr
#> 1 19.74286 17.25132 22.23439
#> 2 15.10000 13.33822 16.86178
#> 
#> $se.fit
#>         1         2 
#> 1.2182168 0.8614094 
#> 
#> $df
#> [1] 29
#> 
#> $residual.scale
#> [1] 3.223099

get_predicted(mod, newdata = nd)
#> Error in mm %*% vcovmat: non-conformable arguments


traceback()

5: get_predicted_se(x, predictions, data = data, ci_type = ci_type,
       vcov_estimation = vcov_estimation, vcov_type = vcov_type,
       vcov_args = vcov_args)
4: get_predicted_ci.default(x, predictions, data = args$data, ci_type = args$ci_type,
       ...)
3: get_predicted_ci(x, predictions, data = args$data, ci_type = args$ci_type,
       ...)
2: get_predicted.lm(mod, newdata = nd)
1: get_predicted(mod, newdata = nd)
```

DominiqueMakowski:
Thanks @vincentarelbundock for testing and helping to improve this (relatively) new function, much appreciated :)

My hunch is that the error originates from somewhere around here:

https://github.com/easystats/insight/blob/1834558a7d702706981252e8e2cca6de9521c2c1/R/get_predicted_ci.R#L226-L237

I'm not sure how to easily fix it

strengejacke:
The error occurs exactly here:

[...]

PR #427, opened 2021-09-16
get_predicted_ci does not drop factor

vincentarelbundock:
This PR illustrates one of the strategies described in this thread: https://github.com/easystats/insight/issues/423

The problem is that feeding a `terms` to the `model.matrix` function drops factor levels if they are not present in the `data` argument. Feeding it a model instead fixes the issue.

`hurdle` and `zeroinfl` models would still fall prey to this trap, because there is some code to drop offsets that is `terms`-specific.

``` r
library(insight)

mod <- lm(mpg ~ factor(cyl), data = mtcars)

nd <- structure(list(cyl = structure(2:3, .Label = c("4", "6", "8"), class = "factor")), out.attrs = list( dim = c(cyl = 2L), dimnames = list(cyl = c("cyl=6", "cyl=8"))), class = "data.frame", row.names = c(NA, -2L))

nd
#>   cyl
#> 1   6
#> 2   8

get_predicted(mod, newdata = nd) |> as.data.frame()
#>   Predicted        SE   CI_low  CI_high
#> 1  19.74286 1.2182168 17.25132 22.23439
#> 2  15.10000 0.8614094 13.33822 16.86178
```

### DIFF
--- a/R/get_predicted_ci.R
+++ b/R/get_predicted_ci.R
@@ -227,18 +227,23 @@ get_predicted_ci.mlm <- function(x, ...) {
     # else, model.matrix below fails, e.g. for log-terms
     attr(data, "terms") <- NULL
 
-    # model terms, required for model matrix
-    model_terms <- tryCatch(
-      {
-        stats::terms(x)
-      },
-      error = function(e) {
-        find_formula(x)$conditional
-      }
-    )
-
-    # drop offset from model_terms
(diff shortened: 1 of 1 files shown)


strengejacke:
@DominiqueMakowski @bwiernik could you review this PR? looks good to me at a first glance.

bwiernik:
Looks good to me once checks pass

DominiqueMakowski:
looks good to me too!

PR #429, opened 2021-09-17
`requiet` makes `testthat` log readable

vincentarelbundock:
This PR is mainly for aesthetics. Feel free to ignore if you don't think this is a good idea.

The `insight` package is different from most because the test suite must load a bunch of different packages. Many of those print messages on `require`, which makes the `testthat` log basically unreadable.

This PR introduces a helper function called `requiet` (isn't this cute?!) which requires packages but suppresses startup messages:

```r
requiet <- function(package) {
  suppressPackageStartupMessages(
    require(package, warn.conflicts = FALSE, character.only = TRUE)
  )
}
```

This is WORK IN PROGRESS; DO NOT MERGE. I want to look at what the `testthat` log looks like on Github actions before moving forward with this.

### DIFF
--- a/tests/testthat/helper-requiet.R
+++ b/tests/testthat/helper-requiet.R
@@ -1,0 +1,5 @@
+requiet <- function(package) {
+  suppressPackageStartupMessages(
+    require(package, warn.conflicts = FALSE, character.only = TRUE)
+  )
+}
--- a/tests/testthat/test-BayesFactorBF.R
+++ b/tests/testthat/test-BayesFactorBF.R
@@ -1,6 +1,6 @@
 .runThisTest <- Sys.getenv("RunAllinsightTests") == "yes"
 
-if (.runThisTest && require("testthat") && require("insight") && require("stats") && require("BayesFactor")) {
+if (.runThisTest && requiet("testthat") && requiet("insight") && requiet("stats") && requiet("BayesFactor")) {
   x <- correlationBF(y = iris$Sepal.Length, x = iris$Sepal.Width)
   test_that("get_data", {
     expect_true(is.data.frame(get_data(x)))
--- a/tests/testthat/test-FE-formula.R
+++ b/tests/testthat/test-FE-formula.R
@@ -1,4 +1,4 @@
-if (require("testthat") && require("insight")) {
+if (requiet("testthat") && requiet("insight")) {
   gfe <- insight:::.get_fixed_effects
 
   test_that(".get_fixed_effects", {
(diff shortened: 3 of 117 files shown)


bwiernik:
That's a clever idea! (And I always love a good pun function name)

vincentarelbundock:
Cool.

I pushed a few manual tweaks to suppress a couple verbose packages, and now things look much better (to my eye). The log is actually useable!

Removed the WIP tag, because I don't plan to do anything else to this PR. Of course, you should feel free to request changes as necessary and I'll be happy to take a look.

vincentarelbundock:
that is, it looks pretty good except for the `brms`/`stan` stuff on the linux machines. But I've spent my energy to work on this for right now... ;)

strengejacke:
> But I've already spent all the energy I had to work on this for now... ;)

What a pity, I hoped you would do this for the other packages as well... :zany_face:

vincentarelbundock:
Never say "never" :)

strengejacke:
and the snapshot tests are quite verbose.... @IndrajeetPatil can you fix this?

IndrajeetPatil:
> and the snapshot tests are quite verbose

I am not sure what you mean by verbose here. The tests will either run silently (if snapshots match with the output) or will give you a warning that there is a mismatch and ask if you want to update them. Are you referring to these warnings?

strengejacke:
Hm, I thought I saw the "snapshot test added" with all of its output there... but maybe this was only once?

IndrajeetPatil:
Yes, that's true. The first time it creates the snapshots, it will print them to the console, but will be silent thereafter.

strengejacke:
Ah, here they're still shown in the log:
https://github.com/easystats/insight/pull/429/checks?check_run_id=3635323068

This is what we want to get rid off.

bwiernik:
It should not be adding those snapshots during regular runs of the tests. That is indicating that it is generating new snapshots instead of checking against the existing output.

IndrajeetPatil:
Oh, good question. I have no idea to suppress these outputs in the log because GHA are going to produce the snapshots afresh each time it is run. 🤔

bwiernik:
Am I misunderstanding how these tests work? Doesn't that mean they can never fail?

IndrajeetPatil:
Oh, I know what's the problem!

At some point, Daniel deleted the `_snaps` folder in `testthat` folder, which contained the snapshots, and then we forgot to add them back in, so currently these tests are generating snapshots each time they are run! I will add them to master.

IndrajeetPatil:
Done! 

Once you incorporate my commit in the `master` branch into this PR and then rerun the actions, the logs should be cleaner. 🤞

strengejacke:
@IndrajeetPatil Snapshots still shown in the log. hm... any ideas?

Issue #432, opened 2021-09-22
`get_predicted(model, predict = "response")`


vincentarelbundock:
This is a discussion thread about the argument value `predict="response"`, related to issues #431 and #310

Currently, this argument value behaves like this:

>  Finally, `"response"` only differs from the previous option for binomial models where it additionally transforms the predictions into the original response's type (for instance, to a factor)

I would argue that this violates the [principle of least astonishment.](https://en.wikipedia.org/wiki/Principle_of_least_astonishment)

"response" is a well-established value for the `type` argument in *many* `predict` methods. For `predict.glm` and most of the ones I know, setting `type="response"` does *not* round and convert probabilities [as is done here.](https://github.com/easystats/insight/blob/master/R/get_predicted.R#L759)

Moreover, the fact that this conversion applies only to binomial model is likely to cause further astonishment, except for users who read carefully to the end of the long paragraph of documentation.

If the goal is to convert probabilities back into categories, then the more common value is: `type="class"`

What do you think? Should `"response"` be removed as one of the officially supported types?

DominiqueMakowski:
"class" and/or "original" could be suitable alternatives I suppose

vincentarelbundock:
One problem with "class" is: what to do if the user says `get_predicted(predict="class")` and expects the same result as `predict(type="class")`?

We want to encourage people to pass any `type` through `predict`, but this raises ambiguity.

vincentarelbundock:
Beyond the naming convention, I also think there is [a bug in the `.get_predict_transform_response` function](https://github.com/easystats/insight/blob/master/R/get_predicted.R#L713):

``` r
library(insight)
mod <- glm(am ~ hp + factor(cyl), family = binomial, data = mtcars)
get_predicted(mod, predict = "response")
# Predicted values:
# 
#  [1] 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0
# 
# NOTE: Confidence intervals, if available, are stored as attributes and can be accessed using `as.data.frame()` on this output.
```

As far as I can tell, this is only call for binomial models, and only supports two levels in the response. Add to that the apparent bug and the naming troubles…

In my humble opinion, we should seriously consider deprecating this option altogether, until someone has a good name, a clear vision, and enough energy to give this more scope.

bwiernik:
Maybe "classification" might be a good api?

DominiqueMakowski:
[...]

Issue #431, opened 2021-09-21
`get_predicted` arguments: `type` vs. `predict`


vincentarelbundock:
I am confused by the use (and interaction between) the `type` and `predict` arguments of `get_predicted` (knowing that `type` is only supported implicitly through the ellipsis). In this GLM model, the default output of `get_predicted` is equivalent to calling `predict` with `type="response"`:

``` r
library(insight)
mod <- glm(am ~ factor(cyl), family = binomial, data = mtcars)

predict(mod, type = "response") |> head(1)
# Mazda RX4 
# 0.4285714
get_predicted(mod) |> head(1)
# Mazda RX4 
# 0.4285714
get_predicted(mod, predict = "expectation") |> head(1)
# Mazda RX4 
# 0.4285714
```

## Problem 1

`get_predicted` seems to accept a `type` argument through the ellipsis, but gives incorrect results:

``` r
get_predicted(mod, type = "response")[1] 
# Mazda RX4 
# 0.6055325
predict(mod, type = "response")[1]
# Mazda RX4 
# 0.4285714
get_predicted(mod, type = "link")[1]
# Mazda RX4 
# 0.4285714
predict(mod, type = "link")[1]
#  Mazda RX4 
# -0.2876821
```

## Problem 2

`get_predicted` behaves different for different *combinations* of the `predict` and `type` arguments:

``` r
get_predicted(mod, predict = "expectation", type = "link")[1]
# Mazda RX4 
# 0.4285714

get_predicted(mod, predict = "expectation", type = "response")[1]
# Mazda RX4 
# 0.6055325

# weirdest result
get_predicted(mod, predict = "response", type = "response")[1]
# Mazda RX4 
#         0
```

## Problem 3

I think the `type` should be accessible. I understand that one of the main goals of `insight` is to make things more uniform, but in many use-cases it is important to have access to the `type` argument directly without the abstraction/convenience layer of the uniform categories created by `insight`. For example, I’m currently working on a packages with two primary function. The first relies on `insight::get_predicted` to get prediction. The second has to call `stats::predict` directly because `get_predicted` is too slow. This means that users (and programmers) must learn the names of prediction types in two “languages”: their modeling package’s and insight’s.

## Feature request

Setting `predict=NULL` overrides the `predict` argument and pushes `type` forward to the `stats::predict` calls:

``` r
get_predicted(mod, predict = NULL, type = "response") == predict(mod, type = "response")[1]
> TRUE
```

strengejacke:
[...]

PR #434, opened 2021-09-23
get_predicted: `type` vs. `predict` disambiguation

vincentarelbundock:
This is a PR to address issue #431  

I included a bunch of new tests, and what feels like "conservative" warnings and errors.

From my perspective, the main thing that still needs to be hashed out is what to do with the current `predict="response"` argument value. In Issue #432, there are some suggestions for different naming convention. At this stage, I would actually advocate removing it altogether for the reasons stated in the other issue. 
 
Any feedback would be much appreciated! I'm happy to make any change you deem necessary.

### DIFF
--- a/R/get_predicted.R
+++ b/R/get_predicted.R
@@ -10,9 +10,9 @@
 #' * `"link"` returns predictions on the model's link-scale (for logistic models, that means the log-odds scale) with a confidence interval (CI).
 #' * `"expectation"` (default) also returns confidence intervals, but this time the output is on the response scale (for logistic models, that means probabilities).
 #' * `"prediction"` also gives an output on the response scale, but this time associated with a prediction interval (PI), which is larger than a confidence interval (though it mostly make sense for linear models).
-#' * `"classification"` only differs from `"predict"` for binomial models where it additionally transforms the predictions into the original response's type (for instance, to a factor).
-#' * Other strings are passed directly to the `type` argument of the `predict` method supplied by the modelling package.
-#' * When `predict = NULL`, alternative arguments such as `type` will be captured by the `...` ellipsis and passed directly to the `predict` method supplied by the modelling package.
+#' * `"classification"` only differs from `"prediction"` for binomial models where it additionally transforms the predictions into the original response's type (for instance, to a factor).
+#' * Other strings are passed directly to the `type` argument of the `predict()` method supplied by the modelling package.
+#' * When `predict = NULL`, alternative arguments such as `type` will be captured by the `...` ellipsis and passed directly to the `predict()` method supplied by the modelling package.
 #' * Notes: You can see the 4 options for predictions as on a gradient from "close to the model" to "close to the response data": "link", "expectation", "prediction", "classification". The `predict` argument modulates two things: the scale of the output and the type of certainty interval. Read more about in the **Details** section below.
 #' @param iterations For Bayesian models, this corresponds to the number of
 #'   posterior draws. If `NULL`, will return all the draws (one for each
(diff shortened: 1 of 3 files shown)


bwiernik:
I think let's go with "classification".

vincentarelbundock:
Thanks! For what it's worth, I also think "classification" is the best of the available options.

I force-pushed an update with the name change.

DominiqueMakowski (approved):
thanks, looks good to me :)

vincentarelbundock:
> thanks, looks good to me :)

And right as you said that, I noticed that the class-specific `get_predicted` methods had not all been updated. Sorry! Pushing the fix in 2 minutes.

R/get_predicted.R
```diff
@@ -6,7 +6,14 @@
...
+#' @param predict string or `NULL`
+#' * `"link"` returns predictions on the model's link-scale (for logistic models, that means the log-odds scale) with a confidence interval (CI).
+#' * `"expectation"` (default) also returns confidence intervals, but this time the output is on the response scale (for logistic models, that means probabilities).
+#' * `"prediction"` also gives an output on the response scale, but this time associated with a prediction interval (PI), which is larger than a confidence interval (though it mostly make sense for linear models).
+#' * `"classification"` only differs from `"predict"` for binomial models where it additionally transforms the predictions into the original response's type (for instance, to a factor).
```
strengejacke, line 13: I think it should be `"prediction"` here-

R/get_predicted.R
```diff
@@ -6,7 +6,14 @@
...
+#' * Other strings are passed directly to the `type` argument of the `predict` method supplied by the modelling package.
+#' * When `predict = NULL`, alternative arguments such as `type` will be captured by the `...` ellipsis and passed directly to the `predict` method supplied by the modelling package.
```
strengejacke, line 15: I prefer adding parens when referring to functions/methods, so you may change to `predict()` here.

R/get_predicted.R
```diff
@@ -634,25 +613,58 @@ get_predicted.faMain <- function(x, data = NULL, ...) {
   # CI
   if (is.null(ci)) ci <- 0
 
+  # check `predict` user-input
+  supported <- c("expectation", "link", "prediction", "classification")
+  if (isTRUE(verbose) && !is.null(predict) && !predict %in% supported) {
+    warning(sprintf('"%s" is not officially supported by the `get_predicted` function as a value for the `predict` argument. It will not be processed or validated, and will be passed directly to the `predict` method supplied by the modeling package. Users are encouraged to check the validity and scale of the results. Set `verbose=FALSE` to silence this warning, or use one of the supported values for the `predict` argument: %s.', predict, paste(sprintf('"%s"', supported), collapse = ", ")))
```
strengejacke, line 619: Please wrap in `format_message()`.

R/get_predicted.R
```diff
@@ -634,25 +613,58 @@ get_predicted.faMain <- function(x, data = NULL, ...) {
...
+  if (is.null(dots$type)) {
+    predict_arg <- predict
+    if (is.null(predict)) {
+      stop("Please supply a value for the `predict` argument.")
+    }
+  } else {
+    if(is.null(predict)) {
+      predict_arg <- dots$type
+    } else {
+      stop('The `predict` and `type` arguments cannot be used simultaneously. The preferred argument for the `get_predicted` function is `predict`. If you need to pass a `type` argument directly to the `predict` method associated with your model type, you must set `predict` to `NULL` explicitly: `get_predicted(model, predict=NULL, type="response")`')
```
strengejacke, line 636: Please wrap in `format_message()`.

strengejacke (changes requested):

strengejacke:
In general, should we have `expected` as alias for `expectation` (and likewise, `predicted` for `prediction`)? At least for me, these shorter forms are more intuitive.

vincentarelbundock:
I'm not a big fan of aliases like this. I think they just introduce complexity to save a couple letters (and all IDEs do completion). I don't have a strong preference for one version or the other (beyond concern for backward compatibility), but I do think we should stick to one canonical way.

strengejacke:
> and all IDEs do completion

Yes, but not for argument-values inside quotes, I think. But I see your point, having too many options might be confusing (or maybe we have them as undocumented options?).

bwiernik:
I agree that we probably shouldn't document too many options/aliases. But I would be okay with silently accepting `expected` and `predicted` as aliases for expectation/prediction

vincentarelbundock:
I pushed a commit with the 4 changes that @strengejacke requested. I think the the Issues I raised and proposed addressing myself should be done in a separate PR. From my perspective this is good to go, but I'm still happy to make any change you want.

FWIW, I also think we should minimize the number of undocumented aliases. It feels safer to be strict about user and to raise informative warnings and errors, rather than do automagical and undocumented changes to user input.

In the current PR, if users call `get_predicted(predict = "expected")`, they will get a warning about unknown argument value.

vincentarelbundock:
To be clear, I don't hold the views expressed above very strongly. None of this feels like a very big deal to me.

DominiqueMakowski:
> But I would be okay with silently accepting expected and predicted as aliases for expectation/prediction

Same

strengejacke:
Thanks for this PR!

PR #438, opened 2021-10-04
bugfix: get_predicted uses newdata

vincentarelbundock:
Fixing a bug that I introduced myself. `get_predicted.default` should call `predict` using the `newdata` argument instead of `data`.


DominiqueMakowski:
thanks!

PR #439, opened 2021-10-04
get_predicted: multinomial/grouped outcomes

vincentarelbundock:
I would love to get input on some design elements described below.

This PR adds support for multinomial or grouped predictors. Consider two models produced by the `nnet::multinom` function:

``` r
library(nnet)
library(MASS)

# binary outcome
m1 <- nnet::multinom(low ~ age + lwt + race + smoke, data = birthwt)

# multinomial outcome
m2 <- nnet::multinom(ftv ~ age + lwt + race + smoke, data = birthwt)
```

When using `predict(type = "class")`, both models generate a vector. When using `predict(type="probs")`, the first model generates a vector but the second a matrix:

``` r
predict(m1, type = "class") |> head()
# [1] 0 0 0 0 0 0
# Levels: 0 1
predict(m2, type = "class") |> head()
# [1] 0 0 0 0 0 0
# Levels: 0 1 2 3 4 6

predict(m1, type = "probs") |> head()
#        85        86        87        88        89        91 
# 0.1648996 0.2174734 0.4236893 0.4092978 0.4331768 0.3498741
predict(m2, type = "probs") |> head()
#            0         1         2          3           4            6
# 85 0.6235402 0.1703559 0.1722631 0.01267403 0.021166773 1.554436e-18
# 86 0.4147791 0.2748552 0.2382574 0.02847018 0.043638105 2.610941e-08
# 87 0.6074009 0.1886751 0.1301803 0.06231978 0.011423895 2.165768e-33
# 88 0.5898206 0.1955748 0.1391058 0.06313327 0.012365574 2.167984e-32
# 89 0.6530309 0.1646664 0.1172860 0.05520435 0.009812382 4.551866e-34
# 91 0.6496718 0.1873155 0.1194621 0.02506273 0.018487765 1.891162e-19
```

With this new PR, I add a `get_predicted.multinom` method which basically only checks the input `type`. I also add a block to the `.get_predicted_out` which reshapes “wide” matrices into “long” data frames.

``` r
get_predicted(m1, predict = "classification") |> head()
# Predicted values:
# 
# [1] "0" "0" "0" "0" "0" "0"
# 
# NOTE: Confidence intervals, if available, are stored as attributes and can be accessed using `as.data.frame()` on this output.

get_predicted(m2, predict = "classification") |> head()
# Predicted values:
# 
# [1] "0" "0" "0" "0" "0" "0"
# 
# NOTE: Confidence intervals, if available, are stored as attributes and can be accessed using `as.data.frame()` on this output.

get_predicted(m1, predict = "expectation") |> head()
#        85        86        87        88        89        91 
# 0.1648996 0.2174734 0.4236893 0.4092978 0.4331768 0.3498741
# 
# NOTE: Confidence intervals, if available, are stored as attributes and can be accessed using `as.data.frame()` on this output.

get_predicted(m2, predict = "expectation") |> head()
# Predicted values:
# 
#   Row Group Predicted
# 1   1     0 0.6235402
# 2   2     0 0.4147791
# 3   3     0 0.6074009
# 4   4     0 0.5898206
# 5   5     0 0.6530309
# 6   6     0 0.6496718
# 
# NOTE: Confidence intervals, if available, are stored as attributes and can be accessed using `as.data.frame()` on this output.
```

There is an important design decision to make with respect to the output format of predicted values for multinomial or grouped outcomes. As you can see above, I decided to go with a long data frame with two new columns. “Row” gives the row number of the `data` input (or of the dataset used to fit the model; this is important for users to be able to “match” predictions to their data. The “Group” column gives the “level” or “group” of the prediction (here: factor levels of the outcome variable in a multinomial logit model). The main alternative to this strategy would be to present results in a “wide” format similar to the matrix output of `predict`.

Disadvantages of my approach:

  - Slightly memory inefficient: the Row and Group variables are not strictly necessary
  - If users want to merge their data back in, they’ll need to create a `Row` column

Benefit:

  - Future proofing: In this example, we have not computed standard errors or confidence intevals. But if we had them and the data were in wide format, we would need to use a ton of weirdly-named, non-tidy, columns: “Predicted\_1”, “SE\_1”, “Predicted\_2”, “SE\_2”.

For me, the benefit overweighs the disadvantages by a considerable margin, but I'd be very curious to know what you think.


bwiernik:
Great!

Regarding wide format with multiple columns, we do already have a convention (`2_Predicted`, `2_SE`, `2_CI_low`) https://easystats.github.io/easystats/articles/conventions.html

I can see the argument for a wide format, but that quickly becomes too unwieldy for more than 2 responses or groups. Another advantage of the long format is that it is easier to work with while plotting. 

Regarding the column names, `Row` should maybe be `Row_ID` and should be rownames if they are present, not only row numbers. 

`Group` seems a little too specific for me as a label. How about `Level`? This would pair well if the model were also multivariate, when we could also have a `Response` variable to indicate which response.

vincentarelbundock:
Cool.

I'm not wedded to `Group` at all.

FWIW, calling `parameters::parameters` on a `multinom` model creates a `Response` column with level values. Maybe this should not be dispositive, but consistency could be nice. Can we think of models where `predict` produces a matrix, but where "Response" would be inappropriate to describe column identifiers in that matrix?

bwiernik:
Thanks for checking that. Let's be consistent with parameters. The tricky situation I'm thinking about would be a multinomial model with multiple response variables, such as simultaneously predicting vital_status = c("alive", "dead", "nearly dead") and main_course = c("chicken", "fish"). I'm not sure we need to be concerned about that at the moment.

vincentarelbundock:
Got it.

I force-pushed a change "Group" -> "Response"

bwiernik:
It looks like the tests for your changes are failing. Could you check those?

vincentarelbundock:
Ooof, well this took forever to figure out. I think this assignment/reuse pattern was accepted by `testhat` but is no longer valid in version 3:

```r
pred <- expect_warning(.....)
expect_true(.... pred ....)
```

Closing the computer for the night, but it looks like the tests are heading in the right direction.

vincentarelbundock:
FYI: Tests finished running over night and are giving us the green light.

PR #442, opened 2021-10-05
New model-specific methods

vincentarelbundock:
I've been doing some testing with various model types. This PR includes various methods I wrote to handle corner cases where default methods do not work.

For non obvious cases, I included a comment to explain why we need the method in question.

Each new method is accompanied by at least one test.


DominiqueMakowski:
that looks good to me, will merge tomorrow if Daniel doesn't have time to review 

<sub>you might wanna join the easystats org just to get free push to the repo - and don't you worry about having too much power, we (at least me 😁) often push much less tested and clean commits straight to master and break everything ^^</sub>

vincentarelbundock:
> that looks good to me, will merge tomorrow if Daniel doesn't have time to review

Thanks!

> you might wanna join the easystats org just to get free push to the repo

Hmm, that sounds like a good idea. If you'll have me, I promise to avoid pushing controversial/breaking changes without 3rd party review.

bwiernik:
This looks great. Thanks for including the comments for future us to remember why different classes need their own methods

strengejacke:
> tests were wrapped in a massive if(){}

image

Never underestimate the power of the <strike>dark side</strike> base R!

Issue #454, opened 2021-10-19
get_predicted(): formally support 'disp' and 'zi'


bwiernik:
We should formally support `predit = 'disp'` and `predict = 'zi'` and related predictions. This is fairly easy for *glmmTMB* and is possible, but more difficult for *brms*. 

Does *brms* provide facilities to extract the posterior predictive distribution for `sigma` or `zi` responses? @paul-buerkner @ASKurz

DominiqueMakowski:
that sounds good

bwiernik:
For *brms*, we could create disp/zi prediction functionality using `hypothesis()` https://paul-buerkner.github.io/brms/articles/brms_distreg.html

strengejacke:
should be resolved in #501 to #503

PR #461, opened 2021-11-03
Don't warn if get_predicted() is called with a model-supported predic…

bwiernik:
…t type


R/get_predicted.R
```diff
@@ -950,7 +950,10 @@ get_predicted.faMain <- function(x, data = NULL, ...) {
   if (is.null(ci)) ci <- 0
 
   # check `predict` user-input
-  supported <- c("expectation", "expected", "link", "prediction", "predicted", "classification")
+  supported <- c(
+    c("expectation", "expected", "link", "prediction", "predicted", "classification"),
+    eval(formals(getS3method("predict", class(x)))$type)
```
strengejacke, line 955: should be `utils::getS3method` here.

bwiernik:
Hmm, the current behavior is just not a great situation.

Despite what the warning says, a non-supported `predict` is not passed to `type`. It just becomes `"response"`. 

I think what we need to do is to check if `predict` is one of the *easystats* expected keywords. If it is, then go through the easystats get_predicted() and get_predicted_ci() workflow. If not, then we just pass the arguments to `predict()` and return whatever it gives back. What do you think @strengejacke ?

``` r
m <- lm(mpg ~ hp + disp, data = mtcars)
predict(m, type = "terms")
#>                              hp       disp
#> Mazda RX4            0.91132048  2.1461456
#> Mazda RX4 Wag        0.91132048  2.1461456
#> Datsun 710           1.33360186  3.7241520
#> Hornet 4 Drive       0.91132048 -0.8277895
#> Hornet Sportabout   -0.70328480 -3.9231097
#> Valiant              1.03552088  0.1736376
#> Duster 360          -2.44209048 -3.9231097
#> Merc 240D            2.10364437  2.5497511
#> Merc 230             1.28392170  2.7287941
#> Merc 280             0.58839942  1.9155139
#> Merc 280C            0.58839942  1.9155139
#> Merc 450SE          -0.82748520 -1.3679533
#> Merc 450SL          -0.82748520 -1.3679533
#> Merc 450SLC         -0.82748520 -1.3679533
#> Cadillac Fleetwood  -1.44848723 -7.3218927
#> Lincoln Continental -1.69688804 -6.9577374
#> Chrysler Imperial   -2.06948926 -6.3508119
#> Fiat 128             2.00428405  4.6132979
#> Honda Civic          2.35204519  4.7043367
#> Toyota Corolla       2.02912413  4.8439296
#> Toyota Corona        1.23424153  3.3569620
#> Dodge Challenger    -0.08228277 -2.6485661
#> AMC Javelin         -0.08228277 -2.2237183
#> Camaro Z28          -2.44209048 -3.6196470
#> Pontiac Firebird    -0.70328480 -5.1369608
#> Fiat X1-9            2.00428405  4.6041940
#> Porsche 914-2        1.38328202  3.3508928
#> Lotus Europa         0.83680023  4.1156189
#> Ford Pantera L      -2.91405202 -3.6499933
#> Ferrari Dino        -0.70328480  2.6013397
#> Maserati Bora       -4.67769778 -2.1326794
#> Volvo 142E           0.93616056  3.3296504
#> attr(,"constant")
#> [1] 20.09062
as.data.frame(insight::get_predicted(m, predict = "terms"))
#> Warning: "terms" is not officially supported by the `get_predicted()` function as a value
#>   for the `predict` argument. It will not be processed or validated, and will be
#>   passed directly to the `predict()` method supplied by the modeling package.
#>   Users are encouraged to check the validity and scale of the results. Set
#>   `verbose=FALSE` to silence this warning, or use one of the supported values for
#>   the `predict` argument: "expectation", "link", "prediction", "classification".
#>                     Predicted        SE    CI_low  CI_high
#> Mazda RX4            23.14809 0.6434873 21.832012 24.46417
#> Mazda RX4 Wag        23.14809 0.6434873 21.832012 24.46417
#> Datsun 710           25.14838 0.7840090 23.544900 26.75186
#> Hornet 4 Drive       20.17416 0.8627647 18.409604 21.93871
#> Hornet Sportabout    15.46423 0.8897264 13.644536 17.28393
#> Valiant              21.29978 0.7624018 19.740497 22.85907
#> Duster 360           13.72542 0.9802259 11.720638 15.73021
#> Merc 240D            24.74402 0.9283932 22.845243 26.64280
#> Merc 230             24.10334 0.7062419 22.658914 25.54777
#> Merc 280             22.59454 0.6245428 21.317205 23.87187
#> Merc 280C            22.59454 0.6245428 21.317205 23.87187
#> Merc 450SE           17.89519 0.6166748 16.633945 19.15643
#> Merc 450SL           17.89519 0.6166748 16.633945 19.15643
#> Merc 450SLC          17.89519 0.6166748 16.633945 19.15643
#> Cadillac Fleetwood   11.32025 1.3786814  8.500525 14.13997
#> Lincoln Continental  11.43600 1.2522950  8.874769 13.99723
#> Chrysler Imperial    11.67032 1.1030980  9.414235 13.92641
#> Fiat 128             26.70821 0.9032742 24.860804 28.55561
#> Honda Civic          27.14701 0.9633679 25.176698 29.11732
#> Toyota Corolla       26.96368 0.9239526 25.073983 28.85337
#> Toyota Corona        24.68183 0.7463262 23.155420 26.20824
#> Dodge Challenger     17.35978 0.8245026 15.673479 19.04607
#> AMC Javelin          17.78462 0.7508830 16.248896 19.32035
#> Camaro Z28           14.02889 0.9892941 12.005554 16.05222
#> Pontiac Firebird     14.25038 1.1264369 11.946557 16.55420
#> Fiat X1-9            26.69910 0.9026091 24.853060 28.54515
#> Porsche 914-2        24.82480 0.7520220 23.286742 26.36286
#> Lotus Europa         25.04304 0.8949892 23.212586 26.87350
#> Ford Pantera L       13.52658 1.1628190 11.148348 15.90481
#> Ferrari Dino         21.98868 1.1102113 19.718043 24.25932
#> Maserati Bora        13.28025 2.2033982  8.773792 17.78670
#> Volvo 142E           24.35644 0.7561244 22.809988 25.90288
```

<sup>Created on 2021-11-03 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

bwiernik:
It's not just 'terms' though. If someone asks for any `type` we don't support, like `"zilink"` in glmmTMB, we shouldn't silently return "response" values

strengejacke:
I think what needs to be changed are these lines:
https://github.com/easystats/insight/blob/de9411a75317725f0b5dabba12af3fb9de789d58/R/get_predicted.R#L1012-1015

`type_arg` is what is used for `type` later. Not sure of we need to set `scale` accordingly

strengejacke:
> I think what needs to be changed are these lines: https://github.com/easystats/insight/blob/de9411a75317725f0b5dabba12af3fb9de789d58/R/get_predicted.R#L1012-1015
> 
> `type_arg` is what is used for `type` later. Not sure of we need to set `scale` accordingly

@DominiqueMakowski can you help here?

DominiqueMakowski:
If I understand correctly yes, currently `type_arg` is hardcoded, but if a `type` argument is passed directly it should overwrite (and become) `type_arg`

bwiernik:
I think it should be--if `predict` is one of the easystats_supported options, it should go through the current type_arg, get_predicted_CI workflow. But if predict is supplied with another option or type is supplied, then the whole workflow needs to be skipped and just call `predict()`

strengejacke:
> If I understand correctly yes, currently `type_arg` is hardcoded, but if a `type` argument is passed directly it should overwrite (and become) `type_arg`

But that never happens. We _only_ have `type = "response"` for Gaussian, and `type = "link"` for all other models when calling `predict()`.

DominiqueMakowski:
but the goal is to allow stuff like `zi_link` isn't it?

strengejacke:
Yes, I guess related to #454. Currently, `type` (if provided) is only used to modulate `ci_type` and `scale`, but not `type_arg` (see https://github.com/easystats/insight/pull/461#issuecomment-960518033). The problem is we only return predictions on the response or link scale, but it becomes more complicated for models with different components. Not sure at which places in the code we need to make changes if we allow more `type` types.

bwiernik:
The argument we use is `predict`. We don't have a user facing argument called `type`. The behavior should be that if `type` is given (as part of `...`) and `predict` is `NULL`, the get_predicted process should be skipped and the arguments just passed to the predict() method instead

vincentarelbundock:
> The argument we use is `predict`. We don't have a user facing argument called `type`. The behavior should be that if `type` is given (as part of `...`) and `predict` is `NULL`, the get_predicted process should be skipped and the arguments just passed to the predict() method instead

One issue to be careful about this is that confidence intervals will be incorrect if we just go straight to the `predict()` methods (e.g., when `predict=NULL` and `type="response"). This is because by default predictions are made on the link scale and the confidence intervals are then transformed based on the information in `get_predicted_args`.

strengejacke:
ok, now, if `type` is provided, it should just be passed to `predict()` "as is". No transformation etc.

vincentarelbundock:
> ok, now, if `type` is provided, it should just be passed to `predict()` "as is". No transformation etc.

Just to be clear, does this mean I can no longer get correctly scales confidence intervals with 

```r
get_predicted(mod, predict=NULL, type="response")
```

A lot of my package assumes this...

strengejacke:
I'm not quite sure about the implementation / usage of the `type_arg`, `scale`, `transform` and `predict` elements returned by `.get_predicted_args()`. Maybe @DominiqueMakowski can provide some information on this? If `type = "response"` is a special case (looks like that to me), we could also change it internally to `"link"` and then set `transform` to true, which would return CI/SE on the correct scale for `type="response"` for all models.

vincentarelbundock:
> I'm not quite sure about the implementation / usage of the `type_arg`, `scale`, `transform` and `predict` elements returned by `.get_predicted_args()`. Maybe @DominiqueMakowski can provide some information on this? If `type = "response"` is a special case (looks like that to me), we could also change it internally to `"link"` and then set `transform` to true, which would return CI/SE on the correct scale for `type="response"` for all models.

Frankly, I can't remember it at all, and I spent several hours with the code. Maybe what I really need is not access to the `type` argument, but a comprehensive dictionary mapping every argument accepted by the various models' `predict()` methods to the names that `insight::get_predicted` has assigned them.

Also, if there's no "true" support for type, we should probably take some steps to make things safer, such as never including confidence intervals.

strengejacke:
@bwiernik I think the warning is misleading. `predict`, if not supported, isn't actually directly passed to `predict()`, see following code in `.get_predicted_args()`:

```r
    # Type (that's for the initial call to stats::predict)
    if (predict_arg == "terms") {
      type_arg <- "terms"
    } else if (info$is_linear) {
      type_arg <- "response"
    } else {
      type_arg <- "link"
    }
```

@vincentarelbundock `get_predicted()` actually has only two official options: either predictions on the link, or on the response scale, for the "full" model (i.e. for ZI models, the predictions taking the ZI component into account). We now wanted to _add_ the possibility to use the `type` argument, which is directly passed to `predict()`, and thereby offering the full range of predict-types that each class-specific `predict()` provides.

bwiernik:
That's what I've been trying to say. 

It _should_ be directly passed. That is what we thought it would do when the warning was written. 

What needs to happen here is that if type or predict aren't in the easystats_supported set, the whole process of this function to get predict args/predicted/ci needs to be bypassed.

vincentarelbundock:
IMHO, it would also be *GREAT* if the next person who figures out all the relationships between `predict`, `type`, `scale`, `transform`, etc. could give them longer, more explicit names. That might make it easier to find our way through the code.

strengejacke:
> That's what I've been trying to say.
> 
> It _should_ be directly passed. That is what we thought it would do when the warning was written.
> 
> What needs to happen here is that if type or predict aren't in the easystats_supported set, the whole process of this function to get predict args/predicted/ci needs to be bypassed.

But we cannot completely bypass it, due to https://github.com/easystats/insight/pull/461#issuecomment-968341542

vincentarelbundock:
> But we cannot completely bypass it, due to [#461 (comment)](https://github.com/easystats/insight/pull/461#issuecomment-968341542)

That's precisely the reason why my previous PR did not bypass it.

But perhaps it would be OK to just be strict about *not* including confidence intervals at all in types we don't support officially.

strengejacke:
ok, let's try to summarize what we would expect:

1. if `predict` is an easystats-supported type (`c("expectation", "link", "prediction", "classification")`), we have the normal processing and either have predictions on the link or response scale, CIs are confidence or prediction intervals
2. if `predict = NULL`, and `type = "response"`, we want predictions on the response scale, including CI _on the same scale_
3. if `predict = NULL`, and `type` is something else, we want predictions on the scale defined in `type`, and CIs possibly as well on that scale. No transformation to predictions and CI here, right?
4. if `predict` is _not_ an easystats-supported type, we would treat it as if `type` was provided instead. Then proceed either with 3. or 4.

Additionally to consider: if we have `"response"` (or `"expectation"`), we need to change the scale to `"link"` and set `tranformation = TRUE`.

vincentarelbundock:
Sounds good, but I would note that 

(3) seems a bit dangerous, since our confidence intervals are typically added by `get_predicted_ci` rather than an option from the modeling package's `predict` method.

(4) sounds good, but doesn't it obviate the need for the convoluted `predict=NULL`? I initially relied on this because it sounded like you wanted to be super strict about what the `predict` argument accepts (there were lots of `match.arg` calls). But if `predict` accepts anything, then we can just  raise a warning along the lines of: `type not supported, use predict instead`. Then, proceed as in your (3) and (4) steps.

bwiernik:
If you can wait a few days, I can do some work on this later this week or next

strengejacke:
> but doesn't it obviate the need for the convoluted `predict=NULL`? I initially relied on this because it sounded like you wanted to be super strict about what the `predict` argument accepts (there were lots of `match.arg` calls).

I'm not sure, but I think we started being more relaxed here when discussing #454. Maybe we should indeed rewrite the code so that we only allow the current options for `predict`, and if `type` is provided, "translate" this if necessary. Actually, I think that only `transfomation` might be affects when `type` is used.

strengejacke:
> If you can wait a few days, I can do some work on this later this week or next

Yes, it's not that urgent (not sure, though, how this interacts with _modelbased_ or _marginaleffects_).

vincentarelbundock:
> Yes, it's not that urgent (not sure, though, how this interacts with _modelbased_ or _marginaleffects_).

Should be a very easy fix on my end. No worries.

strengejacke:
@DominiqueMakowski I think we should consider a revision of `.get_predicted_args()`. 

- The goal should be to have a safe way to define `args$type`, depending on the `predict` argument, i.e. checking the `predict` argument
- if `type` is specified via the ellipses should be used as `args$type`, this should be used (and `predict` ignored, possibly throwing a warning if both are specified)
- check exceptions where `arg$scale` is actually needed. I think this is currently only for one model class, so we might think defining this value independent from `ci_type`?

strengejacke:
`get_predicted.bife` is the only method where we use `args$scale` instead of `args$type`

strengejacke:
This last commit is WIP, will continue on it later...

strengejacke:
``` r
library(insight)
library(ggeffects)
library(glmmTMB)
library(pscl)

data(Salamanders)

m1 <- zeroinfl(count ~ mined | mined, dist = "poisson", data = Salamanders)

ggpredict(m1, "mined", type = "fe") |> as.data.frame()
#>     x predicted  std.error conf.low conf.high group
#> 1 yes  1.398481 0.14247709 1.057742  1.848984     1
#> 2  no  3.553978 0.03831048 3.296892  3.831110     1
get_predicted(m1, data = get_datagrid(m1, at = "mined"), predict = "count") |> as.data.frame()
#>   Predicted         SE   CI_low  CI_high
#> 1  1.398481 0.14247709 1.057742 1.848984
#> 2  3.553978 0.03831048 3.296892 3.831110

ggpredict(m1, "mined", type = "zero_inflated") |> as.data.frame()
#>     x predicted  std.error conf.low conf.high group
#> 1 yes 0.2954545 0.06172082 0.176789  0.414120     1
#> 2  no 2.2648810 0.12824157 2.021319  2.508444     1
get_predicted(m1, data = get_datagrid(m1, at = "mined"), predict = "expectation") |> as.data.frame()
#>   Predicted         SE    CI_low   CI_high
#> 1 0.2954545 0.06444986 0.1660294 0.4248796
#> 2 2.2648810 0.13414830 1.9997385 2.5300236

ggpredict(m1, "mined", type = "zi_prob") |> as.data.frame()
#>     x predicted  std.error  conf.low conf.high group
#> 1 yes 0.7887318 0.14247709 0.7384732 0.8315354     1
#> 2  no 0.3627194 0.03831048 0.3455478 0.3802485     1
get_predicted(m1, data = get_datagrid(m1, at = "mined"), predict = "zero") |> as.data.frame()
#>   Predicted         SE    CI_low   CI_high
#> 1 0.7887318 0.14247709 0.7384732 0.8315354
#> 2 0.3627194 0.03831048 0.3455478 0.3802485
```

<sup>Created on 2022-01-29 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

I fixed some issues with ZI/hurdle models from _pscl_, and we can get predictions for different components of those models (see also #454). I think we probably can in general allow all supported `type` values for the `predict` argument, to be more flexible. Our "main" values are expectation etc., but for certain model classes, we also allow "special" types. What do you say?

vincentarelbundock:
> I think we probably can in general allow all supported `type` values for the `predict` argument, to be more flexible. Our "main" values are expectation etc., but for certain model classes, we also allow "special" types. What do you say?

I agree. And if any value is accepted in `predict`, then there are opportunities for code simplification, removing all the conflict resolution with `predict=NULL, type=???`. I can look into it if you want.

strengejacke:
If we agree, I suggest merging this PR and open a new one where we revise `.get_predicted_args()`, where we copy values of `type` to `predict`, if only `type` is provided. Then we only (need to) take care of `predict`'s values.

strengejacke:
``` r
library(pscl)
library(glmmTMB)
library(insight)

data("fish")
m1 <- glmmTMB(
  count ~ child + camper,
  ziformula = ~ child + camper,
  data = fish,
  family = poisson()
)

m2 <- zeroinfl(
  count ~ child + camper | child + camper,
  data = fish,
  dist = "poisson"
)

predict(m1, type = "response") |> head()
#> [1] 2.536481 7.286753 2.536481 2.095147 2.536481 0.471689
predict(m2, type = "response") |> head()
#>         1         2         3         4         5         6 
#> 2.5364822 7.2867527 2.5364822 2.0951474 2.5364822 0.4716893

get_predicted(m1) |> head()
#> [1] 2.536481 7.286753 2.536481 2.095147 2.536481 0.471689
get_predicted(m2) |> head()
#> [1] 2.5364822 7.2867527 2.5364822 2.0951474 2.5364822 0.4716893



m1 <- glmmTMB(
  count ~ child + camper,
  ziformula = ~ child + camper,
  data = fish,
  family = truncated_poisson()
)

m2 <- hurdle(
  count ~ child + camper | child + camper,
  data = fish,
  dist = "poisson"
)

predict(m1, type = "response") |> head()
#> [1] 2.499998 7.363812 2.499998 2.036180 2.499998 0.420583
predict(m2, type = "response") |> head()
#>         1         2         3         4         5         6 
#> 2.5140014 7.3639287 2.5140014 2.0480429 2.5140014 0.4620445

get_predicted(m1) |> head()
#> [1] 2.499998 7.363812 2.499998 2.036180 2.499998 0.420583
get_predicted(m2) |> head()
#> [1] 2.5140014 7.3639287 2.5140014 2.0480429 2.5140014 0.4620445
```

<sup>Created on 2022-01-30 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

PR #466, opened 2021-11-12
switch to poorman

IndrajeetPatil:
cf. https://github.com/easystats/easystats/issues/168

Also
- spell check
- style
- reknit README


strengejacke:
@IndrajeetPatil can you take a look at this?
https://github.com/easystats/insight/runs/4193558264?check_suite_focus=true

IndrajeetPatil:
Yes, already on it:
https://github.com/nathaneastwood/poorman/issues/94

strengejacke:
ah, cool! but nonetheless, if used conditionally, there should be no error at all, so the issue is still something to be addressed from our side.

strengejacke:
Line 22 in display.rmd needs to be in an else-statement

PR #467, opened 2021-11-12
Fix get_variances for glmmTMB models with no random effects

bwiernik:


strengejacke:
@bwiernik maybe you can look at this code:
https://github.com/easystats/insight/blob/b27d181bbb5e09867c1a8686dd41de41fcc60c3e/R/compute_variances.R#L703

Not sure if at the end the mean is the appropriate approach to calculate the single number for the variance. Do you have any thoughts on this?

bwiernik:
Yes, the is a reasonable choice

Issue #471, opened 2021-11-19
`get_predicted.brmsfit` does not work with multinomial or cumulative models


vincentarelbundock:
`get_predicted.brmsfit` does not appear to support multinomial or cumulative models:

``` r
library(brms)
library(insight)
mod <- brm(rating ~ treat + period + (1 | subject), family = cumulative(), data = inhaler,
           silent = 2, backend = "cmdstanr")

head(predict(mod))
#      P(Y = 1) P(Y = 2) P(Y = 3) P(Y = 4)
# [1,]  0.83875  0.15200  0.00675  0.00250
# [2,]  0.82900  0.16100  0.00775  0.00225
# [3,]  0.84700  0.14475  0.00700  0.00125
# [4,]  0.83275  0.15525  0.00800  0.00400
# [5,]  0.82825  0.16025  0.00875  0.00275
# [6,]  0.83000  0.16250  0.00625  0.00125

get_predicted(mod)
# Error in t.default(draws): argument is not a matrix
```

Leaving this here to document a reproducible example. I can probably work on this at some point in the coming days/weeks.

bwiernik:
cf.

```r
tidybayes::add_fitted_draws(inhaler, mod)
```

bwiernik:
The `predict()` output doesn't seem too reasonable to me--there aren't any uncertainty intervals. 

For `"expectation"`, two options seem reasonable:

1. estimated probability and uncertainty for each category

```r
new_inhaler <- data.frame(product = c(-.5, .5), 
                          order = 0, 
                          carryover = 0, 
                          subject = NA)

est_probs <- tidybayes::add_fitted_draws(
  newdata = new_inhaler, model = mod
) |> 
  group_by(product, .category) |> 
  summarize(
    Median = median(.value), CI = .95,
    hdi(.value), .groups = "drop"
  )
format_table(est_probs)
```
```
  product .category   Median       95% CI
1   -0.50         1     0.54 [0.11, 0.93]
2   -0.50         2     0.42 [0.08, 0.66]
3   -0.50         3     0.03 [0.00, 0.14]
4   -0.50         4 6.83e-03 [0.00, 0.08]
5    0.50         1     0.76 [0.35, 1.00]
6    0.50         2     0.23 [0.00, 0.56]
7    0.50         3 6.20e-03 [0.00, 0.06]
8    0.50         4 1.08e-03 [0.00, 0.02]
```

[...]

PR #473, opened 2021-11-25
WIP: get_predicted.brmsfit supports categorical dv

vincentarelbundock:
Fix for https://github.com/easystats/insight/issues/471

TODO:

- [ ] `predict="categorical"`
- [x] fix bootstrap bug I just introduced

### DIFF
--- a/R/get_predicted.R
+++ b/R/get_predicted.R
@@ -554,7 +554,6 @@ get_predicted.glmmTMB <- function(x,
     )
   }
 
-
   # "expectation" for zero-inflated? we need a special handling
   # for predictions and CIs here.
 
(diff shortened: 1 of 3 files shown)


R/get_predicted.R
```diff
@@ -1225,7 +1222,27 @@ get_predicted.faMain <- function(x, data = NULL, ...) {
...
+    # 3rd dimension of the array is the response level. This stacks the draws into:
+    # Rows * Response ~ Draws
+    iter_stacked <- apply(iter, 1, c)
+    predictions <- data.frame(
+      # rows repeated for each response level
+      Row = rep(1:ncol(iter), times = dim(iter)[3]),
+      # response levels repeated for each row
+      Response = rep(dimnames(iter)[[3]], each = dim(iter)[2]),
+      Predicted = apply(iter_stacked, 1, centrality_function)
+    )
```
strengejacke, line 1238: Just to be safe, not sure if `Response` can be a character, should we add `stringsAsFactors = FALSE`?

vincentarelbundock: Good call. It's only been a 1.5 years, but I often forget that `stringsAsFactors` was once `TRUE` by default. Force pushed.

PR #475, opened 2021-11-26
closes #474

### LINKED ISSUE #474: introduce `standardize_column_order()`
In `{parameters}`, we are [repeating ourselves a ton](https://github.com/easystats/parameters/search?q=col_order) while reordering columns. 

We can avoid this by introducing a function like `standardize_column_order()`, which would standardize column ordering.

The function signature would be similar to `standardize_names()`:

```r
standardize_column_order(
  data,
  style = c("easystats", "broom"),
  ...
)
```

This way, if someone has adopted `{broom}` naming conventions, they would still be able to standardize column order for their dataframe.

If you think this is a good idea, I can work on it.

IndrajeetPatil:
- style NEWS and code
- spell check


IndrajeetPatil:
The only issue I see is that if the column in the dataframe is not present in the column order vector, that column is just left out.

What can I change to make sure that that doesn't happen?

``` r
library(insight)

df <- cbind.data.frame(
  CI_Low      = -2.873,
  t           = 5.494,
  CI_High     = -1.088,
  p           = 0.00001,
  Parameter   = -1.980,
  CI          = 0.95,
  df          = 29.234,
  Method      = "Student's t-test"
)

standardize_column_order(df, style = "easystats")
#>   Parameter   CI           Method     t     df     p
#> 1     -1.98 0.95 Student's t-test 5.494 29.234 1e-05
```

<sup>Created on 2021-11-26 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

bwiernik:
It looks like the column order list is pretty comprehensive. In your example, the data frame columns have the CI_ capitalization wrong. Is ignoring case the only case you are concerned or is it broader? Shouldn't we just ensure that the column order list is comprehensive ?

IndrajeetPatil:
My worry is that if someone uses this function and have some columns that we haven't anticipated (let's say because `{broom}` has added some new columns to their outputs), those columns will just be left out.

So I am wondering if a more cautious approach would be to reorder all the columns found in the list in the specified order, and still retain whatever is not found at the end.

bwiernik:
That makes sense. For that, switch to `union(intersect(col_order, names(df)), names(df)`

IndrajeetPatil:
Works like a charm! Thanks

``` r
library(insight)

df <- cbind.data.frame(
  CI_Low      = -2.873,
  t           = 5.494,
  CI_High     = -1.088,
  p           = 0.00001,
  Parameter   = -1.980,
  CI          = 0.95,
  df          = 29.234,
  Method      = "Student's t-test"
)

standardize_column_order(df, style = "easystats")
#>   Parameter   CI           Method     t     df     p CI_Low CI_High
#> 1     -1.98 0.95 Student's t-test 5.494 29.234 1e-05 -2.873  -1.088
```

<sup>Created on 2021-11-27 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

IndrajeetPatil:
What do you think, @strengejacke?

strengejacke:
Looks good to me!

Issue #480, opened 2022-01-05
clean_names not work when bSplines (from the package splines2) are used in (g)lm models


gbiele:
Hello,

first thanks for making this package available!

I guess this is of interest to @strengejacke, who seems to have written the code.

I stumbled over what I think might be a bug when I use the `get_data` command on a fit object from a glm that uses the `bSpline` function from the splines2 package.

I think I also have a solution for the bug.

First, here is a reproducible example
```
library(splines2)
library(insight)

dt = 
  data.frame(
  y = rnorm(100) > 0,
  x = rnorm(100)
)

fit = glm(y ~ bSpline(x),
          family = binomial(),
          dt)

get_data(fit)
```

which is results in the error message

> Error in \`[.data.frame`(md, , needed.vars, drop = FALSE) : 
  undefined columns selected

I think the reason is that the function `.prepare_get_data` creates a vector `needed.vars` in which the variable for splines is incorrectly named. The variable names is, for the reproducible example above, "bSpline(x)" instead of "x", so that one gets an error if one tries to get this variable from the model data `md`. ([here](https://github.com/easystats/insight/blob/99c89b47bcf2ce5457d320a2e678137be4ac2e88/R/utils_get_data.R#L170))

The reason for the wrong variable name seems to be that the `clean_names` function or more specifically the function `.remove_pattern_from_names` function does not have "bSpline" among the checked patterns. (see [here](https://github.com/easystats/insight/blob/65768115bf8362be48a8d091bd8fc541f0d2c06a/R/clean_names.R#L118))

What do you think?

PS: Everything works fine for the coxph models, because no further processing is done for survival models: see [here](https://github.com/easystats/insight/blob/99c89b47bcf2ce5457d320a2e678137be4ac2e88/R/utils_get_data.R#L162)

PR #481, opened 2022-01-05
add bSpline to patterns in .remove_pattern_from_names

gbiele:
This solves https://github.com/easystats/insight/issues/480: clean_names not work when bSplines (from the package splines2) are used in (g)lm models


strengejacke:
Thanks, looks good!

PR #484, opened 2022-01-07
.is_empty_object should not use is.list()

vincentarelbundock:
`.is_object_empty` checks `is.list(x)` and then compacts it if true. The problem is that a `data.frame` is a list, so we compact all data frames by default, which can be very expensive.

This small change can speed things up a bit in some cases, for e.g., with `get_data`

```r
library(lme4)
library(insight)

N <- 1e6
dat <- data.frame(
    x = rnorm(N),
    k = sample(rnorm(50), N, replace = TRUE),
    m = sample(rnorm(10000), N, replace = TRUE))
dat$y <- dat$x + rnorm(dat$k) + rnorm(dat$m) + rnorm(nrow(dat))
mod <- lmer(y ~ x + (1 | k) + (1 | m), data = dat)
```

Old version:

```r
> bench::mark(get_data(mod), max_iterations = 3)
# A tibble: 1 × 13
  expression      min median `itr/sec` mem_alloc `gc/sec` n_itr  n_gc total_time
  <bch:expr>    <bch> <bch:>     <dbl> <bch:byt>    <dbl> <int> <dbl>   <bch:tm>
1 get_data(mod) 12.5s  12.5s    0.0801     225MB    0.881     1    11      12.5s
```

New version:

```r
> bench::mark(get_data(mod), max_iterations = 3)
# A tibble: 1 × 13
  expression      min median `itr/sec` mem_alloc `gc/sec` n_itr  n_gc total_time
  <bch:expr>    <bch> <bch:>     <dbl> <bch:byt>    <dbl> <int> <dbl>   <bch:tm>
1 get_data(mod) 4.25s  4.25s     0.236     181MB     2.12     1     9      4.25s
```

### DIFF
--- a/R/helper_functions.R
+++ b/R/helper_functions.R
@@ -42,8 +42,12 @@
 .is_empty_object <- function(x) {
   if (inherits(x, "data.frame")) {
     x <- as.data.frame(x)
-  }
-  if (is.list(x) && length(x) > 0) {
+    if (nrow(x) > 0 && ncol(x) > 0) {
+      x <- x[!sapply(x, function(i) all(is.na(i)))]
+      x <- stats::na.omit(x) # faster than checking each row and indexing
+    }
+  # a list but not a data.frame
+  } else if (is.list(x) && length(x) > 0) {
     x <- tryCatch(
       {
         .compact_list(x)
@@ -52,16 +56,10 @@
         x
       }
     )
-  }
-  if (inherits(x, "data.frame")) {
-    if (nrow(x) > 0 && ncol(x) > 0) {
-      x <- x[!sapply(x, function(i) all(is.na(i)))]
-      x <- x[!apply(x, 1, function(i) all(is.na(i))), ]
-      # need to check for is.null for R 3.4
-    }
   } else if (!is.null(x)) {
     x <- stats::na.omit(x)
   }
+  # need to check for is.null for R 3.4
   length(x) == 0 || is.null(x) || isTRUE(nrow(x) == 0) || isTRUE(ncol(x) == 0)
 }
 


bwiernik:
Perhaps let's just switch the order of the list and data.frame sections so that we can be more robust?

```r

# is object empty?
.is_empty_object <- function(x) {
  if (inherits(x, "data.frame")) {
    x <- as.data.frame(x)
      x <- x[!sapply(x, function(i) all(is.na(i)))]
      x <- x[!apply(x, 1, function(i) all(is.na(i))), ]
    }
  }
  else if (is.list(x) && length(x) > 0) {
    x <- tryCatch(
      {
        .compact_list(x)
      },
      error = function(x) {
        x
      }
    )
  }
  else if (!is.null(x)) {
    x <- stats::na.omit(x)
  }
  # need to check for is.null for R 3.4
  length(x) == 0 || is.null(x) || isTRUE(nrow(x) == 0) || isTRUE(ncol(x) == 0)
}
```

vincentarelbundock:
Yep, that works. I also changed the naive rowwise indexing to `stats::na.omit`, which can be a fair bit faster. `get_data` is now about 20x in that specific example:

# A tibble: 1 × 13
  expression      min median `itr/sec` mem_alloc `gc/sec` n_itr  n_gc total_time
  <bch:expr>    <bch> <bch:>     <dbl> <bch:byt>    <dbl> <int> <dbl>   <bch:tm>
1 get_data(mod) 564ms  564ms      1.77     139MB     1.77     1     1      564ms
# … with 4 more variables: result <list>, memory <list>, time <list>, gc <list>

vincentarelbundock:
Note that 20x may not seem like much for a function like this, but `get_data` is often 75% of compute time in my `marginaleffects` package!

bwiernik:
Looks good to me one the windows checks pass

bwiernik:
Did you close by accident?

vincentarelbundock:
I rebased. It's in master now.

strengejacke:
`na.omit()` is faster, but doesn't give the desired result:

``` r
x <- data.frame(a = c(1, NA, 3),
                b = c(NA, NA, NA),
                c = c(2, NA, NA))
x
#>    a  b  c
#> 1  1 NA  2
#> 2 NA NA NA
#> 3  3 NA NA

  x <- x[!sapply(x, function(i) all(is.na(i)))]
  x <- x[!apply(x, 1, function(i) all(is.na(i))), ]
  # need to check for is.null for R 3.4
}
x
#>   a  c
#> 1 1  2
#> 3 3 NA

x <- data.frame(a = c(1, NA, 3),
                b = c(NA, NA, NA),
                c = c(2, NA, NA))

  x <- x[!sapply(x, function(i) all(is.na(i)))]
  x <- stats::na.omit(x) # faster than checking each row and indexing
}
x
#>   a c
#> 1 1 2
```

<sup>Created on 2022-01-08 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

strengejacke:
we could profile against `complete.cases()`, maybe that's faster than `apply()`?

vincentarelbundock:
> `na.omit()` is faster, but doesn't give the desired result (3rd row not empty):

Oops, this is embarassing. I pushed a fix and will benchmark different approaches.

vincentarelbundock:
@strengejacke, we could potentially use `is.na` + `rowSums` and gain some more speed if we only compute a flag instead of re-assigning the whole dataset:

```r
N <- 1e5
f <- function(...) ifelse(rbinom(N, 1, .7) == 1, NA, rnorm(N))
dat <- as.data.frame(lapply(1:5, f))
dat <- setNames(dat, paste0("X", 1:5))

fun_apply <- function(df) {
    df <- df[, !sapply(df, function(i) all(is.na(i)))]
    df <- df[!apply(df, 1, function(i) all(is.na(i))), ]
    return(df)
}

fun_sum <- function(df) {
    df <- df[!sapply(df, function(i) all(is.na(i)))]
    df <- df[rowSums(!is.na(df)) > 0,]
    return(df)
}

fun_sum_no_assign <- function(df) {
    df <- df[, !sapply(df, function(i) all(is.na(i)))]
    flag <- any(rowSums(!is.na(df)) > 0)
    return(flag)
}

bench::mark(fun_apply(df), fun_sum(df), fun_sum_no_assign(df), check = FALSE)

  expression                 min   median `itr/sec` mem_alloc `gc/sec` n_itr
  <bch:expr>            <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl> <int>
1 fun_apply(df)         168.96ms 189.08ms      5.38    21.6MB     26.9     3
2 fun_sum(df)            11.84ms  14.95ms     63.8    18.55MB     37.9    32
3 fun_sum_no_assign(df)   3.01ms   3.39ms    232.      8.77MB     55.9   116
```

strengejacke:
Yes, sounds good! Could you take care of this?

vincentarelbundock:
https://github.com/easystats/insight/pull/485

PR #500, opened 2022-01-30
test refactor: truncreg

vincentarelbundock:
I volunteer to clean up the test suite. 3 proposed changes:

1. The `requiet` function that we use to load packages silently now includes a call to `testthat::skip_if_not_installed`. This eliminates the need for massive `if{}` checks at the start of each test file.
2. Group model-specific tests in "packages" files named with this convention: `test-pkg-truncreg.R`. This allows us to easily distinguish the package-specific tests from those who have to do with `insight` functions.
3. Consolidate tests in fewer `test_that` call. When a test fail, we get the line number and the actual command, so we don't actually need a `test_that{}` label every single time. This would make the tests much easier to read and navigate. 

Compare the exiting file:

https://github.com/easystats/insight/blob/master/tests/testthat/test-truncreg.R

To my proposed change:

https://github.com/vincentarelbundock/insight/blob/27e984902e2df94b5157a0ef339e9ae47390a9d2/tests/testthat/test-pkg-truncreg.R


vincentarelbundock:
Sorry for the false promise. I did a few and this would take too long.

strengejacke:
Yeah, that's ok ;-)

Issue #499, opened 2022-01-30
revise `get_predicted()`


strengejacke:
I currently find it particular difficult to follow when predictions are on response scale, when (back-)transformed etc. I suggest revising the underlying functions in a way that

- we always calculate prediction on the link-scale when `predict` is `"expectation"` / `"response"`, so we always need to back-transform.
- CIs are ~ predictions +/- 1.96 * se, on the link scale / scale of linear predictor
- the function to back-transform is _always_ called, and for linear, the link-inverse function is just `"mu"`, so no problem here. we back-transform predictions and CI, and delta-se
- we can then take special cases into account, like zi-prob for glmmTMB etc.

we may streamline the process a bit, to have a simpler API / internal code.

strengejacke:
@vincentarelbundock @bwiernik @DominiqueMakowski what do you think?

vincentarelbundock:
I agree that the process could be streamlined a bit. However, I'm a bit scared by the "default to backtransform" approach. Are we 100% *sure* that we have the capacity to do this accurately for a wide enough variety of models to treat it as a default option? Do we have tests for all of those models and known benchmark results against which to compare our output?

If we can only be 100% confident & tested for `glm()`, `glmmTMB()`, then we should of course do those. But unless we have tests and benchmarks for a given model, then I'd argue that it's best to rely on the developer of the package, even if the output is not as fully-featured (perhaps issuing a note or warning when appropriate).

strengejacke:
Most packages' `predict()` returns the SE on the linear scale, so we need to back-transform anyway...

bwiernik:
I share some of Vincent's concerns. I am also not sure how universally 
~ predictions +/- 1.96 * se, on the link scale / scale of linear predictor will yield accurate intervals. I'd like to incorporate this workflow into models on case by case basis

vincentarelbundock:
> Most packages' `predict()` returns the SE on the linear scale, so we need to back-transform anyway...

Sure. My point is just that "most packages" is not specific enough. If any of the supported models don't return SE on the linear scale, then we risk producing incorrect results unless we have specifically flagged them.

Basically, my view is that `.default` methods are perfectly fine if all we are doing is extracting some information from an object and if we document the output as "raw" info. However, `.default` methods should never be used to apply any kind of transformation.

But maybe I'm paranoid...

strengejacke:
But this is what we already do:
[...]

Issue #451, opened 2021-10-11
`get_predicted.glmmTMB(predict="expectation")` in zero-inflated models


vincentarelbundock:
`get_predicted.glmmTMB(model, predict="expectation")` does not produce the same results as `predict.glmmTMB(model, type="response")` for zero inflated models.

The documentation in the `predict` method states:

> Denoting mu as the mean of the conditional distribution and ‘p’ as the zero-inflation probability.
> “response” expected value; this is mu\*(1-p) for zero-inflated models and ‘mu’ otherwise

Our own method only takes `mu`:

``` r
library(glmmTMB)
library(insight)

data("fish")
m1 <- glmmTMB(
  count ~ child + camper + (1 | persons),
  ziformula = ~ child + camper + (1 | persons),
  data = fish,
  family = truncated_poisson())

get_predicted(m1, predict = "expectation") |> head()
# [1] 1.023045 2.130428 1.023045 1.663721 1.023045 2.633594

predict(m1, type = "response") |> head()
# [1] 0.2400130 0.9726394 0.2400130 0.4613930 0.2400130 0.6362097
```

We can get the same results by applying the formula in the `glmmTMB` docs

``` r
pred <- get_predicted(m1, predict = "expectation") * (1 - predict(m1, type = "zprob"))

head(pred)
# [1] 0.2400130 0.9726394 0.2400130 0.4613930 0.2400130 0.6362097
```

Should we check if the model is zero inflated and do this correction automatically?

bwiernik:
Yes, expectation should incorporate the zero inflation part. 

As you are working through these, can you make a list of options aside from link/expectation/prediction/classification that various models support? (Eg, conditional response)

bwiernik:
I can investigate in the next few weeks

bwiernik:
A few first thoughts:

1. Zero inflated and hurdle models don't have clean likelihoods. They Franken-models with 2 different likelihoods stapled together. That means they don't have a clean link function. 
2. I wonder what the zeroinfl package does for its intervals? (If it has them)
3. Given (1), I think the 2 main options would be (a) assume normality and use the delta method SEs returned from glmmTMB; or (b) use a parametric bootstrap for it. A third option (c) might be to adapt methods from merTools, but not sure if that's feasible given major differences in the two estimators. 
4. This is one of the main reasons to use MCMC for stuff...

strengejacke:
See section _Details_ here: https://strengejacke.github.io/ggeffects/reference/ggpredict.html
[...]

PR #501, opened 2022-01-30
Update get_predicted.R

strengejacke:
See #499


strengejacke:
@vincentarelbundock @bwiernik @DominiqueMakowski in `.get_predicted_args()`, can we combine the two steps "1. step: define prediction type (that's for the initial call to stats::predict)" and "2. step: define target scale of predictions and CI type"? I would say that this is also the place where we add exceptions like all the ZI stuff etc.

- `predict` can be one of the four supported values, or any of the specific `predict()` methods `type` values.
- `type_arg` is set accordingly to `predict`
- since we use `type_arg = "link"` for `"predict = expectation"`, we set `transform = TRUE` whenever predictions on the response scale are requested.

The most exciting part is the `"scale"` argument. If we consider glmmTMB with `type = "zprob"`, we would have `predict = "zprob"`, `transform = TRUE` and we should set `scale` also to `"zprob"`, because we need to know the exact scale (which is not link-inverse of the model, but rather `qlogis` for the binomial model of the zero-infl model).

strengejacke:
Ok, the basic idea is that predictions for most model classes can be returned by processing following four steps:

```
  # prepare arguments for predictions (.get_predicted_args())
  # 1. step: predictions
  # 2. step: confidence intervals
  # 3. step: back-transform
  # 4. step: final preparation
```

As a rough template, see `get_predicted.lm()`: https://github.com/easystats/insight/blob/192d2b2c37a737df2c80d3b1b3c4bda63f2fa59c/R/get_predicted.R

The CIs are not taken from the `predict()` function, but (in step 2) rather calculated via `get_predicted_ci()`, which itself calls `get_predicted_se()`, using model-matrix/vcov multiplication stuff. But if we _have_ SEs from `predict()`, we can change the function call step 2, like in `get_predicted.glmmTMB()`: https://github.com/easystats/insight/blob/192d2b2c37a737df2c80d3b1b3c4bda63f2fa59c/R/get_predicted_mixed.R

In either case, all relevant information for the "standard procedure" is processed in `.get_predicted_args()`:
https://github.com/easystats/insight/blob/192d2b2c37a737df2c80d3b1b3c4bda63f2fa59c/R/get_predicted_args.R

Step 3 is flexible, as the function for back-transforming has an argument where we optionally can provide a link-inverse function. This may be required for some special cases like hurdle/zeroinflated, see https://github.com/easystats/insight/blob/192d2b2c37a737df2c80d3b1b3c4bda63f2fa59c/R/get_predicted_zeroinfl.R

strengejacke:
I'm not sure this draft is currently working as expected, so we might first fix all bugs here, then add new model classes.

strengejacke:
(btw, this re-writing of the code should also solve #454, as we allow to process any valid `type` value)

strengejacke:
This PR possibly closes: #499 #413 #451 #454

strengejacke:
Looks like the revisions are fairly stable. Can you review, so we can see if this can be merged?

DominiqueMakowski (approved):
That's massive work! Thanks

R/get_predicted_args.R
```diff
@@ -0,0 +1,217 @@
+# process predict-specific arguments ------------------------------------------
+
+.get_predicted_args <- function(x,
+                                data = NULL,
+                                predict = "expectation",
+                                include_random = TRUE,
+                                include_smooth = TRUE,
+                                ci = 0.95,
+                                newdata = NULL,
```
vincentarelbundock, line 9: Do we need both `data` and `newdata` here?

R/get_predicted_args.R
```diff
@@ -0,0 +1,217 @@
...
+
+  # Get info
+  info <- model_info(x, verbose = FALSE)
+
+  # Data
+  if (!is.null(newdata) && is.null(data)) data <- newdata
+  if (is.null(data)) data <- get_data(x, verbose = verbose)
+
+  # CI
+  if (is.null(ci)) ci <- 0
```
vincentarelbundock, line 51: Can we leave `ci` NULL and then skip all the confidence interval construction steps? It could be useful to get fast predictions without CI for bootstrapping, for example.

R/get_predicted_args.R
```diff
@@ -0,0 +1,217 @@
...
+  predict_method <- lapply(
+    class(x), function(i) {
+      tryCatch(utils::getS3method("predict", i),
+               error = function(e) NULL)
+    }
+  )
+  predict_method <- tryCatch(predict_method[!sapply(predict_method, is.null)][[1]], error = function(e) NULL)
+
+  # check aliases
+  if (predict %in% c("expected", "response")) {
```
vincentarelbundock, line 63: I vote for being stricter and doing less magic. Not a big deal, obviously. Happy to keep this.

R/get_predicted_args.R
```diff
@@ -0,0 +1,217 @@
...
+      'Set `verbose = FALSE` to silence this warning, or use one of the supported values for the `predict` argument:',
+      paste(" ", paste(sprintf('"%s"', setdiff(easystats_methods, c("expected", "predicted"))), collapse = ", "))
+    )
+    warning(msg, call. = FALSE)
+  }
+
+
+  transform <- FALSE
+
+  # type_arg  = what we pass own to type
```
vincentarelbundock, line 99: This comment is so clear. Love it!

R/get_predicted_args.R
```diff
@@ -0,0 +1,217 @@
...
+  # define prediction type (that's for the initial call to stats::predict)
+
+  # terms
+  if (predict == "terms") {
+    type_arg <- "terms"
+    scale_arg <- "terms"
+    transform <- FALSE
+
+    # linear models are always on response scale (there is no other)
+  } else if (info$is_linear) {
```
vincentarelbundock, line 113: Should warn or error if `!predict %in% c("link", "response")`, and not just silently switch.

R/get_predicted_args.R
```diff
@@ -0,0 +1,217 @@
...
+    type_arg <- "response"
+    scale_arg <- "response"
+    transform <- FALSE
+
+    # type = "response" always on link-scale - for later back-transformation
+  } else if (predict %in% c("expectation", "response", "prediction", "classification")) {
+
+    if (inherits(x, c("hurdle", "zeroinfl", "zerotrunc"))) {
+      # pscl: hurdle/zeroinfl and countreg
+      type_arg <- "count"
```
vincentarelbundock, line 123: Should again warn or error. Maybe use `match.arg`

R/get_predicted_args.R
```diff
@@ -0,0 +1,217 @@
...
+      transform <- FALSE
+
+    } else {
+      # unknown / default
+      type_arg <- scale_arg <- predict
+      transform <- FALSE
+    }
+
+    # unknown
+  } else {
```
vincentarelbundock, line 159: Why check if `predict %in% type_methods` above if we're behaving the same way here anyway? Can we afford to be stricter and just reject unknown `predict` values? Are we confident that `type_methods` covers all the relevant values?

R/get_predicted_args.R
```diff
@@ -0,0 +1,217 @@
...
+      if (!smooth %in% names(data) || include_smooth == FALSE) {
+        include_smooth <- FALSE
+        data[[smooth]] <- mean(get_data(x)[[smooth]], na.rm = TRUE)
+      }
+    }
+  }
+
+  # Random
+  # In case include_random is TRUE, but there's actually no random factors in data
+  if (include_random && !is.null(data) && !is.null(x) && !all(find_random(x, flatten = TRUE) %in% names(data))) {
```
vincentarelbundock, line 186: insert warning

vincentarelbundock (commented):
This is super nice and clean. Well done!

R/get_predicted_args.R
DominiqueMakowski: good point

bwiernik:
I will look at this Friday

R/get_predicted_args.R
strengejacke: yeah, response shouldn't be possible here any longer.

strengejacke: no, sorry, here we indeed still can have `"response"` as option, to harmonize, I convert it to `"expectation"`.

R/get_predicted_args.R
strengejacke: I think this already happens before when we check the input in lines 76 to 91

strengejacke:
``` r
x <- rnorm(15)
y <- x + rnorm(15)
insight::get_predicted(lm(y ~ x), predict = "count")
#> Warning: `predict` = "count"` is not officially supported by `get_predicted()`.
#>   `predict` will be passed directly to the `predict()` method for the model and
#>   not validated.
#>   Please check the validity and scale of the results.
#>   Set `verbose = FALSE` to silence this warning, or use one of the supported
#>   values for the `predict` argument:
#>     "expectation", "link", "prediction", "classification"
#> Predicted values:
#> 
#>  [1] -1.83681113  1.19301495  0.65079090  2.30813170 -0.08019471  0.17493910
#>  [7] -1.40530023  0.16909396 -0.52268640 -1.34620601 -0.01982088  0.02949426
#> [13]  0.10453808  0.64631889  0.80421114
#> 
#> NOTE: Confidence intervals, if available, are stored as attributes and can be accessed using `as.data.frame()` on this output.
```

<sup>Created on 2022-02-01 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

R/get_predicted_args.R
strengejacke: This is some magic from @bwiernik:

```r
  # check `predict` user-input
  predict_method <- lapply(
    class(x), function(i) {
      tryCatch(utils::getS3method("predict", i),
               error = function(e) NULL)
    }
  )
  predict_method <- tryCatch(predict_method[!sapply(predict_method, is.null)][[1]], error = function(e) NULL)

  type_methods <- suppressWarnings(eval(formals(predict_method)$type))
```

so yes, `type_methods` should cover all possible types. It is rather that there might be valid `type` values that we did not cover yet, and these should be caught somewhere. We need to extend this if-else list when we add model by model.

R/get_predicted_args.R
strengejacke: What do you exactly mean here?

vincentarelbundock: I can't tell from this chunk of code, but I just wanted to make sure that we warn users whenever the`type` is changed on the fly automatically. If we can always transform it back to the scale the user originally requested, there is no need to include a warning and we can move on. But if we force it to `count` and then convert it when we can -- leaving some unknown cases out -- then we need a warning.

strengejacke:
Support for survreg looks good, but transforned SE not... The CI look correct, though.

``` r
library(survival)
library(insight)

m1 <- survreg(Surv(time, status) ~ ph.ecog + age + strata(sex), data = lung)
d <- get_datagrid(m1, list(ph.ecog = c(0, 1, 2)))

predict(m1, newdata = d, se.fit = TRUE)
#> $fit
#>        1        2        3 
#> 583.7849 422.0409 305.1098 
#> 
#> $se.fit
#>        1        2        3 
#> 63.11499 25.35537 31.25380
get_predicted(m1, data = d) |> as.data.frame()
#> Warning: Could not apply Delta method to transform standard errors.
#>   These are returned on the link-scale.
#>   Predicted         SE   CI_low  CI_high
#> 1  583.7849 0.10811344 472.3093 721.5712
#> 2  422.0409 0.06007799 375.1596 474.7806
#> 3  305.1098 0.10243460 249.6110 372.9483

predict(m1, newdata = d, type = "link", se.fit = TRUE)
#> $fit
#>        1        2        3 
#> 6.369533 6.045102 5.720672 
#> 
#> $se.fit
#>          1          2          3 
#> 0.10811344 0.06007799 0.10243460
get_predicted(m1, data = d, predict = "link") |> as.data.frame()
#>   Predicted         SE   CI_low  CI_high
#> 1  6.369533 0.10811344 6.157634 6.581431
#> 2  6.045102 0.06007799 5.927351 6.162853
#> 3  5.720672 0.10243460 5.519904 5.921440

predict(m1, newdata = d, type = "linear", se.fit = TRUE)
#> $fit
#>        1        2        3 
#> 6.369533 6.045102 5.720672 
#> 
#> $se.fit
#>          1          2          3 
#> 0.10811344 0.06007799 0.10243460
get_predicted(m1, data = d, predict = "linear") |> as.data.frame()
#>   Predicted         SE   CI_low  CI_high
#> 1  6.369533 0.10811344 6.157634 6.581431
#> 2  6.045102 0.06007799 5.927351 6.162853
#> 3  5.720672 0.10243460 5.519904 5.921440

predict(m1, newdata = d, type = "lp", se.fit = TRUE)
#> $fit
#>        1        2        3 
#> 6.369533 6.045102 5.720672 
#> 
#> $se.fit
#>          1          2          3 
#> 0.10811344 0.06007799 0.10243460
get_predicted(m1, data = d, predict = "lp") |> as.data.frame()
#>   Predicted         SE   CI_low  CI_high
#> 1  6.369533 0.10811344 6.157634 6.581431
#> 2  6.045102 0.06007799 5.927351 6.162853
#> 3  5.720672 0.10243460 5.519904 5.921440
```

<sup>Created on 2022-02-01 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

bwiernik:
Where is the code for the survreg method?

strengejacke:
```r
#' @export
get_predicted.survreg <- get_predicted.lm
```

(https://github.com/easystats/insight/pull/501/files#diff-961f97138ccec5adee92e55de344b61feefa2d507869055d8400159ab820b6d9R240-R248)

strengejacke:
I think the issue with the CI / SE for survreg is: SE's on link-scale are correct, thus CIs are "correct" - however, delta-method for transforming SE's fails.

strengejacke:
The new code in `.get_predicted_transform()` now gives a warning, and does not fail when delta-method is failing:

```r
      # Transform SE (https://github.com/SurajGupta/r-source/blob/master/src/library/stats/R/predict.glm.R#L60)
      # Delta method; SE * deriv( inverse_link(x) wrt lin_pred(x) )
      mu_eta <- tryCatch(abs(get_family(x)$mu.eta(predictions)), error = function(e) NULL)
      if (is.null(mu_eta)) {
        if (isTRUE(verbose)) {
          warning(format_message("Could not apply Delta method to transform standard errors.",
                                 "These are returned on the link-scale."), call. = FALSE)
        }
      } else {
        ci_data[se_col] <- ci_data[se_col] * mu_eta
      }
```

bwiernik:
What do you mean by "failing"?

strengejacke:
`get_family(x)$mu.eta` is not available for all model classes.

``` r
library(survival)
library(insight)

m1 <- survreg(Surv(time, status) ~ ph.ecog + age + strata(sex), data = lung)
get_family(m1)$mu.eta
#> NULL
```

<sup>Created on 2022-02-02 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

bwiernik:
What is held in mu.eta?

strengejacke:
I don't know, I haven't added that code. Maybe @DominiqueMakowski ?

DominiqueMakowski:
No recollection of that :/

bwiernik:
Let me look later this week

strengejacke:
> No recollection of that :/

@DominiqueMakowski:

image

But:
https://github.com/easystats/insight/commit/2fbdb56feb6d0e67fd6a3838785893c0473ec6e4#diff-961f97138ccec5adee92e55de344b61feefa2d507869055d8400159ab820b6d9

bwiernik:
I love git blame!

vincentarelbundock:
> I love git blame!

the name is just so great

DominiqueMakowski:
But in my defence, it was simply code taken from base R 😁

strengejacke:
Do we want to merge these 3 PRs?

DominiqueMakowski:
yes!

PR #506, opened 2022-02-05
find_predictors.fixest

vincentarelbundock:
Currently, `get_predictors.fixest` does not extract instruments from multipart formulas.


strengejacke:
Thanks!

PR #514, opened 2022-02-28
WIP: Improve metafor support

bwiernik:
tl;dr: What do you think of my adding `include_interval` arguments to `get_data.rma()` rather than these being in `parameters::parameters()`? @strengejacke @DominiqueMakowski @IndrajeetPatil @mattansb 

I'm looking to improve our handling of meta-analysis objects, particularly from *metafor*. 

The first step here is to get various functions in *insight* working. 

Currently, I've improved missing data handling in `get_data.rma()`, and made `get_formula.rma()` work for most cases.

I have also added arguments to include confidence intervals for effect sizes in `get_data.rma()`. This is to facilitate creation of forest plots. Currently, we have this functionality in `parameters::model_parameters.rma()`, but it doesn't make much sense to me there. The observed study effect sizes are not really model parameters--they are the observed raw data. Including them in `parameters()` also produces odd output for meta-analysis with predictors (vs intercept-only). So I suggest we move this functionality here. I think the reason they might be in `parameters()` is because they are in `broom::tidy()`, but this doesn't make much sense to me there either. This choice was probably to facilitate making forest plots, but there are better ways to do that, and this structure isn't really all that useful for these plots anyway.

Still to come: 

*insight*

- [ ] Update `get_predicted()`

In `get_predicted()`, we should have:
  1. "expectation" : fitted values + confidence interval
  2. "prediction" : fitted values + prediction interval
  3. "blup" : best linear unbiased predictions + CI (analogous to `coef()` or `modelbased::estimate_grouplevel()` in an MLM)

Should the the third option rather be in `parameters::model_parameters(..., group_level = TRUE)`, with options "random" and "total"/"blup"? That would be more similar to how we handled mixed effects models

In addition, `get_predicted()` needs some work to handle *metafor*'s unusual expected inputs to `predict()` (a model matrix rather than a model frame) and unusual output (a single row for intercept-only models).

The various flavors of rma models should be handled.

*parameters*

Update `model_parameters()`:
  - [ ] remove `include_studies`
  - [ ] include dispersion parameters in table

*performance* 

- [ ] Make `check_model()` work

Should generally look and work like with mixed effects models

*modelbased*

- [ ] Check the various `estimate_*()` functions to ensure they work

*see*

- [ ] Add funnel plot functionality
- [ ] Add forest plot functionality
- [ ] Add visualization_recipe() functionality to produce funnel, forest, or scatterplots

I could move this off into their package (*easymeta* or similar) if desired


strengejacke:
You're the meta analysis expert. And in 95% of all cases, the things you did made sense. So I trust you here :-)
In general, your suggestion sounds reasonable to me. I could work on the _parameters_ side of this, but maybe after you finished this PR, so I have a clearer picture of what to do in _parameters_.

DominiqueMakowski:
What Mattan said ;)

arthur-albuquerque:
This would be a very useful commit. Any updates?

strengejacke:
I just saw you wanted to remove `include_studies` and add a funnel plot... "why" to both? 
1. Do you want to completely remove estimates for single studies, or just show them always?
2. Funnel plots are already possible, maybe there's room for improvement: https://easystats.github.io/see/articles/parameters.html#model-parameters-of-meta-analysis-models

bwiernik:
I think we should remove the studies option from parameters and use get_data() for that

PR #522, opened 2022-03-06
vcov

vincentarelbundock:
Superseded arguments in `get_predicted()`, `get_predicted_se()`, and `get_predicted_ci()`:

* `vcov_estimation`
* `vcov_type`

Replaced by:

* `vcov`
* `vcov_args`

Implementation notes:

* The first commit of this PR is a test in which I hard-coded standard errors calculated before the code refactor
* I copied over the `.get_vcov()` function from `parameters`

Questions for future development:

* Would it be nice to make the `vcov` and `vcov_args` arguments available in `get_varcov()`?
* Should `parameters` use the new `insight` facilities to extract `vcov` instead of hosting its own?


bwiernik:
> Would it be nice to make the vcov and vcov_args arguments available in get_varcov()?
> Should parameters use the new insight facilities to extract vcov instead of hosting its own?

Yes and Yes

vincentarelbundock:
> Yes and Yes

Cool. I would probably do this in a subsequent PR to keep things manageable this time.

Issue #521, opened 2022-03-06
get_predicted brms improvements


DominiqueMakowski:
I was about to write concerning this: https://github.com/paul-buerkner/brms/issues/1309, so that `get_predicted()` automatically sets `re_formula = NA` if random variables of data are `NA`.  

But it seems that it does it already as it doesn't fail on the newest (GH version) brms. However, I noticed a few possible improvements:

``` r
library(brms)

data <- mtcars
data$cyl <- as.character(data$cyl)

model <- brm(cyl ~ mpg * vs + (1 | carb),
             data = data,
             chains = 6,
             iter = 3000,
             algorithm = "meanfield",
             refresh=0,
             family = categorical(link = "logit", refcat = "4"))
#> Compiling Stan program...
#> Start sampling
#> Warning: Pareto k diagnostic value is 0.97. Resampling is unreliable. Increasing
#> the number of draws or decreasing tol_rel_obj may help.

# Carb = NA
newdata <- data.frame(mpg = 20, vs = 0, carb = NA)
```

```r
pred <- insight::get_predicted(model, newdata = newdata)
#> Error in posterior_epred.brmsfit(x, newdata = args$data, re.form = args$re.form, : formal argument "newdata" matched by multiple actual arguments
```

1. Though our argument is named data, I suspect some users might try using `newdata` if they try to simply replace the functions with datawizard.  We should probably run a check on the ellipsis and see whether there is an argument called newdata that would supersedes a null data one (with or without out a warning).

```r
pred <- insight::get_predicted(model, data = newdata)
pred
#> Predicted values:
#> 
#>   Row Response    Predicted
#> 1   1        4 0.0003576466
#> 2   1        6 0.6150440761
#> 3   1        8 0.3845982773
#> 
#> NOTE: Confidence intervals, if available, are stored as attributes and can be accessed using `as.data.frame()` on this output.
```

It says in the docs that `iterations=NULL` returns all iterations, but here it returns the summary 🤔🤔 
```r
pred <- insight::get_predicted(model, data = newdata, iterations = 10)
#> Warning: Argument 'nsamples' is deprecated. Please use argument 'ndraws'
#> instead.
pred
#> Predicted values:
#> 
#>   Row Response    Predicted
#> 1   1        4 8.581724e-05
#> 2   1        6 7.944995e-01
#> 3   1        8 2.054146e-01
#> 
#> NOTE: Confidence intervals, if available, are stored as attributes and can be accessed using `as.data.frame()` on this output.
```

[...]

PR #523, opened 2022-03-08
get_predicted brms

DominiqueMakowski:
#521

### DIFF
--- a/R/get_predicted.R
+++ b/R/get_predicted.R
@@ -18,7 +18,8 @@
 #'   posterior draws. If `NULL`, will return all the draws (one for each
 #'   iteration of the model). For frequentist models, if not `NULL`, will
 #'   generate bootstrapped draws, from which bootstrapped CIs will be computed.
-#'   Iterations can be accessed by running `as.data.frame()` on the output.
+#'   Iterations can be accessed by running `as.data.frame(..., keep_iterations = TRUE)`
+#'   on the output.
 #' @param include_random If `"default"`, include all random effects in the
 #'   prediction, unless random effect variables are not in the data.  If `TRUE`,
 #'   include all random effects in the prediction (in this case, it will be
@@ -88,6 +89,11 @@
 #' when provided in `get_predicted()`, are passed down to [get_predicted_ci()],
 #' thus, see the related documentation there for more
 #' details.
+#' }
+#'
+#' \subsection{Bayesian and Bootstrapped models and iterations}{
+#' For predictions based on multiple iterations, for instance in the case of Bayesian models and bootstrapped predictions, the function used to compute the centrality (point-estimate predictions) can be modified via the `centrality_function` argument. For instance, `get_predicted(model, centrality_function = stats::median)`. The default is `mean`.
+#' Individual draws can be accessed by running `iter <- as.data.frame(get_predicted(model))`, and their iterations can be reshaped into a long format by `bayestestR::reshape_iterations(iter)`.
 #' }
 #'
 #' @examples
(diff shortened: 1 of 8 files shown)


R/get_predicted.R
```diff
@@ -480,16 +486,18 @@ get_predicted.bife <- function(x,
...
     )
     iter <- as.data.frame(iter_stacked)
-    names(iter) <- paste0("iter_", names(iter))
     # outcome with a single level
   } else {
     # .get_predicted_boot already gives us the correct observation ~ draws format
     if (is.null(colnames(iter)) || !all(grepl("^iter", colnames(iter)))) {
       iter <- as.data.frame(t(iter))
-      names(iter) <- gsub("^V(\\d+)$", "iter_\\1", names(iter))
+
```
DominiqueMakowski, line 494: This regex wasn't working as expected, giving results like `iter_V1`, `iter_V2` instead of `iter_1`

R/get_predicted_bayesian.R
```diff
@@ -43,7 +43,7 @@ get_predicted.stanreg <- function(x,
   fun_args <- list(x,
                    newdata = args$data,
                    re.form = args$re.form,
-                   nsamples = iterations,
+                   ndraws = iterations,
```
DominiqueMakowski, line 46: This fixes the warning, but I'm not sure wether it might break stuff on very old brms versions 🤔

bwiernik: We can check for which argument is used in the method, ala:

```r
fun_args <- list(
  # arguments other than ndraws
)
if ("nsamples" %in% 
  names(formals(getS3method(
    "posterior_predict",
    class(x)
  )))
) {
  fun_args <- c(
    fun_args, nsamples = iterations
  )
} else {
  fun_args <- c(
    fun_args, ndraws = iterations
  )
}
```

bwiernik: We can check for which argument is used in the method, ala:

```r
fun_args <- list(
  # arguments other than ndraws
)
if ("nsamples" %in% 
  names(formals(getS3method(
    "posterior_predict",
    class(x)
  )))
) {
  fun_args <- c(
    fun_args, nsamples = iterations
  )
} else {
  fun_args <- c(
    fun_args, ndraws = iterations
  )
}
```

R/get_predicted.R
bwiernik: The issue here would be that column names originally look like "iter_V1" rather than "V1", so nothing is being modified because there are no matches.  

The regex should change to 

`gsub("^(?:iter_)*V(\\d+)+", "iter_\\1", names(iter))`

bwiernik (changes requested):
Addressed your comments

R/get_predicted.R
DominiqueMakowski: I wonder if that's more efficient than simply assigning a new vector of names, it's not like we expect anything else than iter_1:ncol()

R/get_predicted_bayesian.R
DominiqueMakowski: why not, but isn't it overkill?

strengejacke: Will the code break at some point, or is it just a warning? If it would be incompatible for older brms/rstanarm versions to use `ndraws`, I would definitely use the suggested code from Brenton, else it's probably a "matter of taste"...

R/get_predicted_bayesian.R
```diff
@@ -46,6 +46,24 @@ get_predicted.stanreg <- function(x,
                    ndraws = iterations,
                    draws = iterations)
 
+  # The following makes the argument passed to predict "nsamples" or "ndraws",
+  # as it got changed in recent brms versions (but we want to preserve compatibility)
+  if ("nsamples" %in%
+      names(formals(getS3method(
```
strengejacke, line 52: `utils::getS3method()`

strengejacke (changes requested):

DominiqueMakowski:
test errors seem to be related to glmmTMB 🤷

PR #529, opened 2022-03-13
new prompt to install packages 

mattansb:
@strengejacke @bwiernik @IndrajeetPatil @DominiqueMakowski 
Any thoughts or feelings about this?


mattansb:
image

image

image

Previous option still works:

``` r
check_if_installed(c("easystats", "ggggggggplot"), prompt = FALSE)
#> Error: Packages 'easystats' and 'ggggggggplot' required for this function to work.
#>   Please install them by running install.packages("easystats", "ggggggggplot").

check_if_installed(c("easystats", "ggggggggplot"), prompt = FALSE, stop = FALSE)
#> Warning: Packages 'easystats' and 'ggggggggplot' required for this function to work.
#>   Please install them by running install.packages("easystats", "ggggggggplot").

check_if_installed(c("easystats", "ggggggggplot"), quietly = TRUE)
```

When not interactive:

```R
check_if_installed(c("easystats", "ggggggggplot")) # when not interactive
#> Error: Packages 'easystats' and 'ggggggggplot' required for this function to work.
#>   Please install them by running install.packages("easystats", "ggggggggplot").
```

R/format_message.R
```diff
@@ -29,8 +29,10 @@ format_message <- function(string, ..., line_length = options()$width) {
     line_length <- 70
   }
 
-  string <- .wrap_message_line(string, line_length)
-  further_lines <- list(...)
+  all_lines <- c(string, ...)
```
strengejacke, line 32: What's the reason behind this change?

strengejacke:
Yes, that's a nice service. I'm usually hesitant regarding automatic updates because in certain cases, R/Rstudio fails because it needs to (re-)start the session, sometimes even a restart of RStudio before you can install updates - it only works after restart, and thus a prompt for installation doesn't help.

IndrajeetPatil:
Looks good to me.

> R/Rstudio fails because it needs to (re-)start the session, sometimes even a restart of RStudio before you can install updates

@strengejacke I have also seen this behavior, but usually for packages that have compiled code.

But if the user needs to install these updates anyway, we might as well give it a try from our side?

Either way, no strong feelings on this one.

strengejacke:
Yeah, I'm ok with that change, or doesn't do any harm anyway 😉

R/format_message.R
mattansb: Else you can't pass a vector as the first argument 🤷‍♂️

mattansb:
> R/Rstudio fails because it needs to (re-)start the session, sometimes even a restart of RStudio before you can install updates

My experience is that this happens when a loaded package is being e-installed.

> But if the user needs to install these updates anyway, we might as well give it a try from our side?

Agree (:

R/check_if_installed.R
```diff
@@ -31,6 +31,7 @@
 #' x
 #' }
 #' @export
+#' @importFrom utils install.packages
```
strengejacke, line 34: No importFrom tags, only namespace

strengejacke (changes requested):

Issue #524, opened 2022-03-11
`get_predicted.default()` `predict` argument


vincentarelbundock:
@strengejacke , could you tell me what the expected behavior is for the `predict` argument of `get_predicted.default()`? I see that this method passes `...` to the `predict()` method, but that it does *not* have a `predict` argument. This means that I can use `type` to change the types of predictions but not `predict`:

``` r
library(survival)
library(insight)
dat <- data.frame(time = c(4, 3, 1, 1, 2, 2, 3),
                  status = c(1, 1, 1, 0, 1, 1, 0),
                  x = c(0, 2, 1, 1, 1, 0, 0),
                  sex = factor(c(0, 0, 0, 0, 1, 1, 1)))
mod <- coxph(Surv(time, status) ~ x + strata(sex),
             data = dat,
             ties = "breslow")

predict(mod, type = "risk")
    ##         1         2         3         4         5         6         7 
    ## 0.4791461 2.0870462 1.0000000 1.0000000 1.6331323 0.7825089 0.7825089

get_predicted(mod) |> head()
    ##          1          2          3          4          5          6 
    ## -0.7357498  0.7357498  0.0000000  0.0000000  0.4904999 -0.2452499

get_predicted(mod, predict = "risk") |> head()
    ##          1          2          3          4          5          6 
    ## -0.7357498  0.7357498  0.0000000  0.0000000  0.4904999 -0.2452499

get_predicted(mod, type = "risk") |> head()
    ##         1         2         3         4         5         6 
    ## 0.4791461 2.0870462 1.0000000 1.0000000 1.6331323 0.7825089
```

I expected that the last two commands would produce the same results. Was I wrong, or should we add a `predict` argument to the default method?

strengejacke:
@DominiqueMakowski Any particular reason to keep this behaviour, or could we support `predict` in the default-method, too?

bwiernik:
Yeah the default should be to pass "predict" to "type" with some translation from our terminology to what's common in other R packages

DominiqueMakowski:
> Any particular reason to keep this behavior

As far I see no, passing predict to type sounds good!

strengejacke:
``` r
library(survival)
library(insight)
dat <- data.frame(time = c(4, 3, 1, 1, 2, 2, 3),
                  status = c(1, 1, 1, 0, 1, 1, 0),
                  x = c(0, 2, 1, 1, 1, 0, 0),
[...]

PR #530, opened 2022-03-14
#524

strengejacke:


strengejacke:
@vincentarelbundock @bwiernik for coxph, `type = "expected"` seems to return incorrect CIs - probably because the back-transformation is not correct. Do you know how to get CIs from the predictions and SEs, when `get_predicted(mod, predict = "expected")`?

From `?predict.coxph`:

> type the type of predicted value. Choices are (...) the expected number of events given the covariates and follow-up time ("expected") (...). The survival probability for a subject is equal to exp(-expected).

PR #534, opened 2022-03-16
get_datgrid: allow for multiple lengths

DominiqueMakowski:


strengejacke:
There are no numeric variables in `at`, and the `for()` loop works different now:

``` r
x <- vector("numeric")
x
#> numeric(0)

for (i in 1:length(x)) {
  print(paste("i =", i))
  print(paste("x[i] =", x[i]))
}
#> [1] "i = 1"
#> [1] "x[i] = NA"
#> [1] "i = 0"
#> [1] "x[i] = "


for (i in x) {
  print(paste("i =", i))
  print(paste("x[i] =", x[i]))
}
```

<sup>Created on 2022-03-16 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

strengejacke:
I think my commit should fix this.

PR #543, opened 2022-03-20
WIP: get_varcov(): vcov and vcov_args arguments

vincentarelbundock:
Currently only works for `get_varcov.default()`

Question/Problem: All `get_varcov.CLASS()` methods accept `...`. This means that when `sandwich` does not support a model, the `vcov` argument will be absorbed silently and return an IID var-cov without any warning.

Do we have to introduce explicit warnings in every single method where `sandwich` doesn't work? What is the most elegant way to do that?

### DIFF
--- a/R/get_predicted_se.R
+++ b/R/get_predicted_se.R
@@ -15,7 +15,7 @@ get_predicted_se <- function(x,
 
   vcovmat <- get_varcov(
     x,
-    vcov_fun = vcov,
+    vcov = vcov,
     vcov_args = vcov_args,
     ...
   )
--- a/R/get_predicted_se.R
+++ b/R/get_predicted_se.R
@@ -13,7 +13,7 @@ get_predicted_se <- function(x,
   # the diagonal of this matrix represent the standard errors of the predictions,
   # which are then multiplied by 1.96 for the confidence intervals.
 
-  vcovmat <- .get_predicted_ci_vcov(
+  vcovmat <- .get_varcov_sandwich(
     x,
     vcov_fun = vcov,
     vcov_args = vcov_args,
@@ -81,133 +81,12 @@ get_predicted_se <- function(x,
 
 
 
-# Get Variance-covariance Matrix ---------------------------------------------------
-
-.get_predicted_ci_vcov <- function(x,
-                                   vcov_fun = NULL,
-                                   vcov_args = NULL,
-                                   verbose = TRUE,
-                                   ...) {
-  dots <- list(...)
-
-  # deprecated
-  if (isTRUE(verbose) && "vcov_type" %in% names(dots)) {
-    warning(format_message("The `vcov_type` argument is superseded by the `vcov_args` argument."), call. = FALSE)
-  }
-  if (isTRUE(verbose) && "robust" %in% names(dots)) {
-    warning(format_message("The `robust` argument is superseded by the `vcov` argument."), call. = FALSE)
-  }
-
-  if (is.null(vcov_args)) {
-    vcov_args <- list()
-  }
-
-  # deprecated: `vcov_estimation`
-  if (is.null(vcov_fun) && "vcov_estimation" %in% names(dots)) {
-    vcov_fun <- dots[["vcov_estimation"]]
-  }
-
-  # deprecated: `robust`
-  if (isTRUE(dots[["robust"]]) && is.null(vcov_fun)) {
-    dots[["robust"]] <- NULL
-    vcov_fun <- "HC3"
-  }
-
-  # deprecated: `vcov_type`
-  if ("vcov_type" %in% names(dots)) {
-    if (!"type" %in% names(vcov_args)) {
-      vcov_args[["type"]] <- dots[["vcov_type"]]
-    }
-  }
-
-  # vcov_fun is a matrix
-  if (is.matrix(vcov_fun)) {
-    return(vcov_fun)
-  }
-
-  # vcov_fun is a function
-  if (is.function(vcov_fun)) {
-    if (is.null(vcov_args) || !is.list(vcov_args)) {
-      args <- list(x)
-    } else {
-      args <- c(list(x), vcov_args)
-    }
-    .vcov <- do.call("vcov_fun", args)
-    return(.vcov)
-  }
-
-  # type shortcuts: overwrite only if not supplied explicitly by the user
-  if (!"type" %in% names(vcov_args)) {
-    if (isTRUE(vcov_fun %in% c(
-      "HC0", "HC1", "HC2", "HC3", "HC4", "HC4m", "HC5",
-      "CR0", "CR1", "CR1p", "CR1S", "CR2", "CR3", "xy",
-      "residual", "wild", "mammen", "webb"
-    ))) {
-      vcov_args[["type"]] <- vcov_fun
-    }
-  }
-
-  # default vcov matrix
-  if (is.null(vcov_fun)) {
-    .vcov <- get_varcov(x, ...)
-    return(.vcov)
-  }
-
-  if (!grepl("^(vcov|kernHAC|NeweyWest)", vcov_fun)) {
-    vcov_fun <- switch(vcov_fun,
-      "HC0" = ,
-      "HC1" = ,
-      "HC2" = ,
-      "HC3" = ,
-      "HC4" = ,
-      "HC4m" = ,
-      "HC5" = ,
-      "HC" = "vcovHC",
-      "CR0" = ,
-      "CR1" = ,
-      "CR1p" = ,
-      "CR1S" = ,
-      "CR2" = ,
-      "CR3" = ,
-      "CR" = "vcovCR",
-      "xy" = ,
-      "residual" = ,
-      "wild" = ,
-      "mammen" = ,
-      "webb" = ,
-      "BS" = "vcovBS",
-      "OPG" = "vcovOPG",
-      "HAC" = "vcovHAC",
-      "PC" = "vcovPC",
-      "CL" = "vcovCL",
-      "PL" = "vcovPL"
-    )
-  }
-
-  # check if required package is available
-  if (vcov_fun == "vcovCR") {
-    check_if_installed("clubSandwich", reason = "to get cluster-robust standard errors")
-    fun <- get(vcov_fun, asNamespace("clubSandwich"))
-  } else {
-    check_if_installed("sandwich", reason = "to get robust standard errors")
-    fun <- try(get(vcov_fun, asNamespace("sandwich")), silent = TRUE)
-    if (!is.function(fun)) {
-      stop(sprintf("`%s` is not a function exported by the `sandwich` package.", vcov_fun))
-    }
-  }
-
-  # extract variance-covariance matrix
-  .vcov <- do.call(fun, c(list(x), vcov_args))
-
-  .vcov
-}
-
 
 # Get Model matrix ------------------------------------------------------------
 
 .get_predicted_ci_modelmatrix <- function(x, data = NULL, vcovmat = NULL, ...) {
   resp <- find_response(x)
-  if (is.null(vcovmat)) vcovmat <- .get_predicted_ci_vcov(x, ...)
+  if (is.null(vcovmat)) vcovmat <- .get_varcov_sandwich(x, ...)
 
 
   if (is.null(data)) {
--- a/R/get_predicted_se.R
+++ b/R/get_predicted_se.R
@@ -13,7 +13,7 @@ get_predicted_se <- function(x,
   # the diagonal of this matrix represent the standard errors of the predictions,
   # which are then multiplied by 1.96 for the confidence intervals.
 
-  vcovmat <- .get_varcov_sandwich(
+  vcovmat <- get_varcov(
     x,
     vcov_fun = vcov,
     vcov_args = vcov_args,
@@ -86,7 +86,7 @@ get_predicted_se <- function(x,
 
 .get_predicted_ci_modelmatrix <- function(x, data = NULL, vcovmat = NULL, ...) {
   resp <- find_response(x)
-  if (is.null(vcovmat)) vcovmat <- .get_varcov_sandwich(x, ...)
+  if (is.null(vcovmat)) vcovmat <- get_varcov(x, ...)
 
 
   if (is.null(data)) {
--- a/R/get_varcov.R
+++ b/R/get_varcov.R
@@ -40,9 +40,12 @@ get_varcov <- function(x, ...) {
   dots <- list(...)
   if ("vcov" %in% names(dots) && !is.null(dots$vcov)) {
     # list of supported classes from Zeileis et al. (2020) JSS
-    supported <- c("lm", "glm", "betareg", "clm", "coxph", "survreg", "crch",
-                   "hurdle", "zeroinfl", "pscl", "countreg", "mlogit", "polr",
-                   "rlm", "fixest")
+    supported_sandwich <- c("lm", "glm", "betareg", "clm", "coxph", "survreg",
+                            "crch", "hurdle", "zeroinfl", "pscl", "countreg",
+                            "mlogit", "polr", "rlm", "fixest")
+    supported_clubSandwich <- c("glm", "gls", "ivreg", "lm", "lme", "lmerMod",
+                                "mlm", "plm", "rma.mv", "rma.uni", "robu")
+    supported <- c(supported_sandwich, supported_clubSandwich)
     if (!any(supported %in% class(x))) {
       msg <- sprintf("Models of class `%s` may not be supported by the `sandwich` or `clubSandwich` packages. In such cases, `get_varcov()` ignores the `vcov` argument and tries to return the model object's default variance-covariance matrix.", class(x)[1])
       warning(format_message(msg), call. = FALSE)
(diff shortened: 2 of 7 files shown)


strengejacke:
> What is the most elegant way to do that?

Maybe just mention in the docs?

vincentarelbundock:
> Maybe just mention in the docs?

I think I'd like to convince the `easystats` crew to be *much* stricter w.r.t. unsupported arguments. Often times it's fine to sink arguments silently in `...`. I do it all the time in `modelsummary` where the worse that can happen is that your table won't be aligned correctly, or somesuch. However, if the arguments are expected to change statistical results, there should at minimum be a warning. Otherwise, people might publish results claiming HC-robust standard errors when they are not. That would be very bad.

vincentarelbundock:
Also, nobody reads the docs (sadly). ;)

strengejacke:
I'm ok with having a warning in each method for classes that are not support by (club)sandwich. I thought your question aimed at avoiding this ;-)

vincentarelbundock:
One possibility is to run a sanity check and raise a warning in the `get_varcov` *generic* rather than in the individual methods. See here:

https://github.com/easystats/insight/pull/543/files#diff-3acbbf81e2521b2920713f77af713a5c06ed09db0e3bcac01bfbce0232f44122L39

I've never done something like this in a package, but it seems to work and I can't immediately think of a disadvantage.

strengejacke:
Ok, looks good to me! But we should include the packages supported by clubSandwich in the checks, too.

bwiernik:
A little annoying that the sandwich/clubSandwich packages don't have their own checks for unsupported models. @vincentarelbundock Do you have a reprex for an unsupported model I could play with?

Some other thoughts:

1. Rather than a warning, which might have a fair number of false positives, do we just want a message() or console output?
2. Another option might be to compare the IID vcov() to the returned robust vcov(). If they are identical, then maybe return a warning in that case?

vincentarelbundock:
I’m not sure there will be a ton of false positives, honestly, so I doubt the warning will be very intrusive. And FWIW, I feel like the risk of unexpected/incorrect numerical results is precisely the kind of thing that should raise a warning rather than just console output (which I see as more appropriate for progress bars and such). I’m even tempted by `error()`, though that probably comes with too many problems. Point is: This is mission critical stuff.

Here’s a reprex:

``` r
library(lme4)
library(sandwich)
mod <- lmer(mpg ~ hp + (1 | cyl), data = mtcars)
vcovHC(mod)
#> Error in UseMethod("estfun"): no applicable method for 'estfun' applied to an object of class "c('lmerMod', 'merMod')"
```

vincentarelbundock:
Just to emphasize, the warning is only triggered if the user explicitly supplied a non-`NULL` value for the `vcov` argument AND if the model is not explicitly supported.

bwiernik:
I'd prefer an `error()` rather than a `warning()`. I agree with you about the mission critical nature, but generally I think that `warning()` is a poor choice--most cases should either be errors or messages. 

I'm confused though. That example doesn't return the IID silently for me. Instead, it gives the error from the vcovHC() function:

``` r
library(lme4)
library(sandwich)
mod <- lmer(mpg ~ hp + (1 | cyl), data = mtcars)
vcovHC(mod)
#> Error in UseMethod("estfun"): no applicable method for 'estfun' applied to an object of class "c('lmerMod', 'merMod')"
insight::get_varcov(mod, vcov = "HC")
#> Warning: Models of class `lmerMod` may not be supported by the `sandwich` or
#>   `clubSandwich` packages. In such cases, `get_varcov()` ignores the `vcov`
#>   argument and tries to return the model object's default variance-covariance
#>   matrix.
#> Error in UseMethod("estfun"): no applicable method for 'estfun' applied to an object of class "c('lmerMod', 'merMod')"
```

<sup>Created on 2022-03-20 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

vincentarelbundock:
> I'd prefer an `error()` rather than a `warning()`. I agree with you about the mission critical nature, but generally I think that `warning()` is a poor choice--most cases should either be errors or messages.

Interesting! I don't have a view on this, so I'm not arguing, just curious: What's annoying about warnings?

> I'm confused though. That example doesn't return the IID silently for me.

Riiight. I have not implemented a `tryCatch` yet. If we go with `error` (which I favor), we will never reach that point anyway, so the `sandwich` error will never show up. If we go with a message I can implement a catch which returns the default vcov.

bwiernik:
It looks like all of the *sandwich* methods rely on `estfun()` S3 dispatch, so we could intercept the errors from sandwich with `tryCatch()` and then return them with our more informative error message. 

*clubSandwich* doesn't use S3 in that way, so not sure that will work reliably for `"CR"` unfortunately. In that case, maybe checking if the output is identical might be preferred? It seems like some non-explicitly supported models might successfully complete `vcovCR.default()`?


Also, typo:

```r
      msg <- sprintf("Models of class `%s` may not be supported by the `sandwich` or `clubSandwich` packages. In such cases, `get_varcov()` ignores the `vcov` argument and tries to return the model object's default variance-covariance matrix.", 
        class(mod)[1])
```

`class(mod)[1]` should be `class(x)[1]`

vincentarelbundock:
> we could intercept the errors from sandwich with `tryCatch()` and then return them with our more informative error message.

*Where* in the code do you think we should catch this. Consider these two stacks:

`get_varcov()` -> `get_varcov.default()` -> `.get_varcov_sandwich()`

`get_varcov()` -> `get_varcov.clm2()` -> `stats::vcov()`

Problems:

1. `sandwich` only gets called at the very last step, so we can't compare or catch before then
2. `sandwich` is never called for unsupported models. If we insert the sanity checks after we can catch (i.e., the *last* step), then we will never error for unsupported models and `vcov` will be silently absorbed. (As I said above, I'm trying to avoid having to modify every single method.)

strengejacke:
@vincentarelbundock Is this PR something you would like to take some time for? I'd like to submit _insight_ the next days, so we might wait for the update after next submission for this PR?

vincentarelbundock:
Yep, probably makes sense to wait a bit for this one. I won't merge it until I see the new version has hit CRAN

bwiernik:
Why not put the sensibility check inside .get_vcov_sandwich() and then don't tryCatch() it or make it error with a specific message that can be extracted inside of .process_vcov() ?

vincentarelbundock:
Because there are ~30 different `get_varcov` methods and only one of them (the default) calls `get_varcov_sandwich`. So the check would almost never be triggered.


The other constraint is that I would like to avoid implementing 30 checks in 30 methods.

vincentarelbundock:
This is why I was trying to find an acceptable sanity check to run *before* calling sandwich

bwiernik:
I don't understand how these arguments can work at all in the other methods then?

vincentarelbundock:
Right, they do NOT work. They are absorbed silently and ignored through the `...` ellipsis. The goal is to raise a warning in those cases where the `vcov` argument is ignored because unsupported.

bwiernik:
Can we take a step back? We _want_ them to work. We need to make the various methods handle the arguments properly or else what's the point of adding them at all?

vincentarelbundock:
Yes we want them to work, but we can only make a method work if the class is supported by `sandwich` or `clubSandwich`.

I feel I didn't express myself clearly. Here's what I propose:

1. Make a list of all the model classes supported by `sandwich` or `clubSandwich`
2. Implement the `vcov` and `vcov_args` arguments for each of those classes. This is done via a combination of `get_varcov.default` and specific`get_varcov.CLASSNAME` methods.
3. Raise a warning (or error) if the user explicitly calls the `vcov` argument and the model object is not of one of the classes in point 1.

In my posts above, I explained why it easier to implement the model class sanity check at the very top of `get_varcov()`,  *before* we call the `sandwich::vcovHC` function.

With that proposal, the user will only get warnings (or errors) if they try to use the `vcov` argument with an unsupported model type.

vincentarelbundock:
@bwiernik I pushed a new commit with a different (cleaner?) strategy:

1. Each `get_varcov` method starts with this call: `.sanity_get_varcov_dots(x, ...)`
2. If a method explicitly supports `vcov` argument in its argument list, then `vcov` will not be pushed forward through `...`
3. If `vcov` is not in `...`, then we know that it is not "officially" supported, and we raise a warning.
4. If `vcov` is officially supported by calling `sandwich` or `clubSandwich` raises an error, we catch it gracefully in `get_varcov_sandwich()` and return an informative error message.

Does that work for you?

bwiernik:
That strategy sounds great. Can we please call it `.check_get_varcov_dots()` instead of `.sanity_get_varcov_dots()`?

R/get_predicted_se.R
```diff
@@ -13,7 +13,7 @@ get_predicted_se <- function(x,
   # the diagonal of this matrix represent the standard errors of the predictions,
   # which are then multiplied by 1.96 for the confidence intervals.
 
-  vcovmat <- .get_predicted_ci_vcov(
+  vcovmat <- .get_varcov_sandwich(
```
strengejacke, line 16: Shouldn't this call `get_varcov()`, and then automatically use sandwich estimators if required, and normal vcov if not?

R/get_predicted_se.R
```diff
@@ -81,133 +81,12 @@ get_predicted_se <- function(x,
...
-  .vcov
-}
-
 
 # Get Model matrix ------------------------------------------------------------
 
 .get_predicted_ci_modelmatrix <- function(x, data = NULL, vcovmat = NULL, ...) {
   resp <- find_response(x)
-  if (is.null(vcovmat)) vcovmat <- .get_predicted_ci_vcov(x, ...)
+  if (is.null(vcovmat)) vcovmat <- .get_varcov_sandwich(x, ...)
```
strengejacke, line 90: See comment above

R/get_varcov.R
```diff
@@ -46,8 +45,30 @@ get_varcov <- function(x, ...) {
...
+                               ...) {
+  .check_get_varcov_dots(x, ...)
+  dots <- list(...)
+  if ("robust" %in% names(dots)) {
+    # default robust covariance
+    if (is.null(vcov)) {
+      vcov <- "HC3"
+    }
+    warning("The `robust` argument is deprecated. Please use `vcov` instead.")
+  }
```
strengejacke, line 61: Please add `call. = FALSE` argument  here.

R/get_varcov.R
```diff
@@ -874,3 +951,13 @@ get_varcov.LORgee <- get_varcov.gee
...
   }
 }
+
+.check_get_varcov_dots <- function(x, ...) {
+  dots <- list(...)
+  # if `vcov` is in ..., it means that the argument is not explicitly
+  # supported by a method, and thus that it will be ignored.
+  if ("vcov" %in% names(dots)) {
+    msg <- sprintf("The `vcov` argument of the `insight::get_varcov()` function is not yet supported for models of class `%s`.", paste(class(x), collapse = "/"))
+    warning(format_message(msg))
```
strengejacke, line 961: See above, please add `call. = FALSE`

strengejacke:
I'm not quite sure how this PR will work. We currently only have sandwich-estimators for the default method, but what is with all the other supported classes, like hurdle/zeroinfl from pscl? These currently don't call `.get_varcov_sandwich()`.

strengejacke:
Furthermore, does `.check_get_varcov_dots()` do anything for the default method? All models not supported by (club)sandwich will not be detected here, because the default method has a `vcov` argument, so here will never be a `vcov` arg in the dots, if I'm not mistaken.

vincentarelbundock:
> I'm not quite sure how this PR will work. We currently only have sandwich-estimators for the default method, but what is with all the other supported classes, like hurdle/zeroinfl from pscl? These currently don't call `.get_varcov_sandwich()`.

Correct. Currently the `vcov` argument is only supported in the default method, which applies to a lot of models, include `lm` and `glm`. If we want to support it elsewhere, we will need to modify the individual methods, such as `get_vcov.hurdle()` to call `.get_varcov_sandwich()`.

Note that this functionality does not exist at all currently, so this is not a regression. We can improve individual methods gradually when we find time or have a need.

> Furthermore, does `.check_get_varcov_dots()` do anything for the default method? All models not supported by (club)sandwich will not be detected here, because the default method has a `vcov` argument, so here will never be a `vcov` arg in the dots, if I'm not mistaken.

That's exactly the point. When `vcov` is explicitly in the argument list of a method, `...` will *not* include `vcov`, and `.check_get_vcov_dots()` will not trigger a warning. In contrast, when `vcov` is not in a method's argument list, it will be pushed to `.check_get_varcov_sandwich` through `...`, and it will then trigger a warning.

In `get_varcov.default`, the `vcov` argument will never trigger a warning, because that function uses `vcov`, so it is effectively useless. I only included it there in `get_varcov.default()` in case we eventually introduce richer checks in that function.

R/get_predicted_se.R
vincentarelbundock: ebe8d2e4

R/get_predicted_se.R
vincentarelbundock: ebe8d2e4

R/get_varcov.R
vincentarelbundock: ebe8d2e4

R/get_varcov.R
vincentarelbundock: ebe8d2e4

strengejacke:
ok, understand!

strengejacke:
I think we can just compare for which classes we also have a `get_varcov()` method, and then include the call to sandwich:

``` r
library(sandwich)
methods("bread")
#>  [1] bread.clm*      bread.coxph*    bread.default*  bread.gam*     
#>  [5] bread.glm*      bread.hurdle*   bread.lm*       bread.mlm*     
#>  [9] bread.mlogit*   bread.nls*      bread.polr*     bread.rlm*     
#> [13] bread.survreg*  bread.zeroinfl*
#> see '?methods' for accessing help and source code

library(clubSandwich)
methods("vcovCR")
#>  [1] vcovCR.default* vcovCR.glm*     vcovCR.gls*     vcovCR.ivreg*  
#>  [5] vcovCR.lm*      vcovCR.lme*     vcovCR.lmerMod* vcovCR.mlm*    
#>  [9] vcovCR.plm*     vcovCR.rma.mv*  vcovCR.rma.uni* vcovCR.robu*   
#> see '?methods' for accessing help and source code
```

All other methods should automatically go into our `get_varcov.default()`.

strengejacke:
turns out, these are not pretty much :-) only mlm needs to be covered, if I made no mistake.

``` r
library(sandwich)
sand1 <- as.character(methods("bread"))

library(clubSandwich)
#> Registered S3 method overwritten by 'clubSandwich':
#>   method    from    
#>   bread.mlm sandwich
sand2 <- as.character(methods("vcovCR"))

library(insight)
gvc <- as.character(methods("get_varcov"))

sand1 <- gsub("(.*)\\.(.*)", "\\2", sand1)
sand2 <- gsub("(.*)\\.(.*)", "\\2", sand2)
gvc <- gsub("(.*)\\.(.*)", "\\2", gvc)

intersect(gvc, unique(c(sand1, sand2)))
#> [1] "default"  "hurdle"   "mlm"      "zeroinfl"
```

<sup>Created on 2022-03-27 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

vincentarelbundock:
@strengejacke not sure what your preferred merge order and release schedule are. Are you taking it from here, or would you like me to do something else?

strengejacke:
First, thanks for this PR, this makes it easy to streamline the robust vcov-stuff!
SInce _insight_ is currently under submission (CRAN team investigating due to a note during submission), I'd like to wait with PRs until I get feedback from CRAN. In case I need to do little changes, I would like not to have "untested" features in the CRAN version to avoid any new issues during submission process. However, this PR and #549 seem to be stable, so we still might merge even when I need to resubmit.

Anyway, TL;DR, usually I would like to wait until CRAN acceptance of _insight_ before merging.

tests/testthat/test-get_varcov.R
```diff
@@ -5,9 +5,12 @@ test_that("informative error in get_varcov.default", {
     requiet("lme4")
     mod <- lmer(mpg ~ hp + (1 | cyl), data = mtcars)
     # sandwich: not supported
-    expect_error(get_varcov(mod, vcov = "HC2"), regexp = "not be supported")
```
strengejacke: @vincentarelbundock Not sure, but can it be that due to line breaks in the warning message, the regex might be different here? To be safe, I just removed the test against the regex.

vincentarelbundock: Oh right. That's totally possible.

vincentarelbundock:
Sounds great. I don't think there's any rush to merge this, so feel free to wait until next release.

However, if you need to tweak something and have a chance to add a default value to `get_predicted(predict)`, I would definitely do that because it's a tiny change, but the current version risks breaking code for people.

strengejacke:
> However, if you need to tweak something and have a chance to add a default value to get_predicted(predict), I would definitely do that because it's a tiny change, but the current version risks breaking code for people.

It's fixed in #549 however, this was after submission, so it will probably not make it into this update.

strengejacke:
Yesterday evening I got the mail about the revdep issues, so submission process is probably almost finished. This means, the three recent PRs in _insight_ will go into the _next_ version and not be included in this current update.

PR #544, opened 2022-03-20
fix failing get_data test

vincentarelbundock:
An attempt to fix the currently failing test on `master` --- UNSUCCESSFUL

### DIFF
--- a/tests/testthat/test-get_data.R
+++ b/tests/testthat/test-get_data.R
@@ -33,6 +33,7 @@ test_that("lm with transformations", {
 })
 
 test_that("mgcv", {
+  skip("works interactively")
   requiet("mgcv")
   d <- iris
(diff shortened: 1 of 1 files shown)


strengejacke:
Yeah, maybe just remove that test? I have also done several attempts, all unsuccessful.

vincentarelbundock:
Yeah, it's a tough one. I have several like that in `marginaleffects`, where the test works interactively but not in `R CMD check`.  This mostly when I call `get_data`, but I don't think it's an actual problem with `get_data`, but rather some weird interaction between environments and how `testthat` creates separate environments to run the tests in.

I added this:

```r
skip("works interactively")
```

At least we'll still have the test in our arsenal, and `testthat` will print an informative message that some tests are skipped because of this.

PR #559, opened 2022-04-17
get_modelmatrix.betareg() supports `data` argument

vincentarelbundock:
`get_modelmatrix` does not currently support the `data` argument, which means that `get_predicted_ci` does not support the `data` argument for `betareg` models.


vincentarelbundock:
A rather simple PR. I don't think the current errors are related to the new additions, so it's probably safe to merge this.

strengejacke:
The data argument is captured in the dots, see https://github.com/vincentarelbundock/insight/blob/13ab02c36d56b5214bd9e2914a2ff1a5107fe182/R/get_modelmatrix.R#L25

vincentarelbundock:
Cool, no problem. I just force pushed an update.

Out of curiosity, why are we catching it through `...`? Wouldn't it be better to be explicit and to document it?

strengejacke:
Honestly, I'm not sure if there was a specific reason. 😁
But we can change it and make it an official argument

vincentarelbundock:
Cool. I guess I'll merge this now and get back to it if I find the energy.

PR #573, opened 2022-06-01
clean_names: correctly extract several variables from mgcv::s()

### LINKED ISSUE #553: Question - Is there a way to find variables for smooth components in mgcv::gam?
Hello! Thanks for this package, it's so great!

I have a question about GAMs with mgcv.

I wonder if there is a function to programmatically find variables based on smooth term strings (without having to regex the string).

For example:

```r
library(mgcv)
dat <- gamSim(1,n=400,dist="normal",scale=2)
b <- gam(y~s(x0)+s(x1)+s(x2)+s(x3),data=dat)
```

there are four smooth terms and I would like to be able to extract the variables in the terms, so that for example from `"s(x0)` I get `"x0"` and so on (in principle regexing would work, but smooth specifications can get so complicated that it's a bit of a puzzle making sure you get indeed the variable).

Is this possible with insight?

etiennebacher:
Close #553

``` r
library(insight)
library(mgcv)
#> Loading required package: nlme
#> This is mgcv 1.8-40. For overview type 'help("mgcv-package")'.

dat <- gamSim(1,n=400,dist="normal",scale=2)
#> Gu & Wahba 4 term additive model
b <- gam(y~x0+s(x1)+s(x2)+x3+s(x1,x2),data=dat)

find_smooth(b, flatten = TRUE) |> clean_names()
#> [1] "x1"     "x2"     "x1, x2"
```

<sup>Created on 2022-06-01 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>


etiennebacher:
Works with `mgcv::s()` but not with `gam::s()`:
``` r
mgcv::s(Age)
#> $term
#> [1] "Age"
#> 
#> $bs.dim
#> [1] -1
#> 
#> $fixed
#> [1] FALSE
#> 
#> $dim
#> [1] 1
#> 
#> $p.order
#> [1] NA
#> 
#> $by
#> [1] "NA"
#> 
#> $label
#> [1] "s(Age)"
#> 
#> $xt
#> NULL
#> 
#> $id
#> NULL
#> 
#> $sp
#> NULL
#> 
#> attr(,"class")
#> [1] "tp.smooth.spec"
gam::s(Age)
#> Error in as.matrix(x): object 'Age' not found
```

<sup>Created on 2022-06-01 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

Using `eval(parse)` is probably the wrong approach because of this namespace conflict

IndrajeetPatil (approved):

etiennebacher:
actually, I noticed some improvements to make, don't merge now

PR #579, opened 2022-06-07
add GHA for codecov

### LINKED ISSUE #577: Add codecov back?
Apparently, `insight` is the only easystats package not covered by codecov (see [here](https://app.codecov.io/gh/easystats)). I see that it was available before but removed in https://github.com/easystats/insight/commit/1d0ccc1e048d5007ff4ce4e8bbf06b161399d75a. Is it possible to put it back to have more info about the package test coverage?

etiennebacher:
Close #577


.github/workflows/test-coverage.yaml
```diff
@@ -0,0 +1,31 @@
...
+  push:
+    branches: [main, master]
+  pull_request:
+    branches: [main, master]
+
+name: test-coverage
+
+jobs:
+  test-coverage:
+    runs-on: ubuntu-latest
```
IndrajeetPatil, line 13: If this fails, then you can try:

```yaml
runs-on: macOS-latest
```

Sometimes there are issues because some deps  can't be installed for Ubuntu.

IndrajeetPatil (approved):

.github/workflows/test-coverage.yaml
IndrajeetPatil: How many and which tests fail? Maybe we can skip them while computing code coverage.

etiennebacher: And yet there is no failure on the master branch... (except for macos-latest with R-devel that is broken)

etiennebacher: That's weird, devtools::test() fails locally but running the failing tests individually work

etiennebacher: @IndrajeetPatil two of them fail, one in test-MCMglmm, the other in test-survreg

etiennebacher:
Strangely, the tests work correctly with `devtools::check()` but not with `devtools::test()`. The problem is that when we run `devtools::test()` the objects loaded in each test file are kept in the global environment for the next tests (which I guess is not the case for `devtools::check()`). This creates some odd bugs sometimes, like the following. In `test-logistf.R`, the package `logistf` is loaded, and for some reason, this makes `test-MCMCglmm` fail:

* when only `insight` and `MCMCglmm` are loaded, it works well:
``` r
library(insight)
library(MCMCglmm)
#> Loading required package: coda
#> Loading required package: ape

data(PlodiaPO)
MCMCglmm(
  PO ~ plate,
  random = ~FSfamily,
  data = PlodiaPO,
  verbose = FALSE,
  nitt = 1300,
  burnin = 300,
  thin = 1
)
#> $Sol
#> Markov Chain Monte Carlo (MCMC) output:
#> Start = 301 
#> End = 1300 
#> Thinning interval = 1 
#>         (Intercept)      plate
#>    [1,]   0.9768808 0.04499501
#>    [2,]   1.0137710 0.04702829
#>    [3,]   1.0100584 0.04153752
#>    [4,]   1.0180808 0.04454613
#>    [5,]   1.0066746 0.04389696
#>    [6,]   1.0274956 0.03688405
#>    [7,]   1.0527792 0.03421160
#>    [8,]   1.0001955 0.04813859
#>    [9,]   1.0610042 0.03441205

[...]

#> 
#> attr(,"class")
#> [1] "MCMCglmm"
```

<sup>Created on 2022-06-07 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

* when `logistf` is loaded, it breaks `MCMCglmm`:
``` r
library(insight)
library(MCMCglmm)
#> Loading required package: coda
#> Loading required package: ape
library(logistf)

data(PlodiaPO)
MCMCglmm(
  PO ~ plate,
  random = ~FSfamily,
  data = PlodiaPO,
  verbose = FALSE,
  nitt = 1300,
  burnin = 300,
  thin = 1
)
#> Error in terms.formula(formula, data = data): invalid term in model formula
```

<sup>Created on 2022-06-07 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

I have no idea why simply loading this package breaks the formula. Regarding the tests, I think codecov uses `devtools::test()`, which is why it fails while R-check works. There should be a line at the end of each test file to clear the environment and the packages loaded. Maybe `on.exit()` could work for this.

strengejacke:
> There should be a line at the end of each test file to clear the environment

We should be careful with that because we define some variables in the initial testthat-file, and the helper-function `requiet()`. not sure if these get lost?

etiennebacher:
Indeed that's probably not the best way to do it, but maybe we could wrap all the code in each file in `withr::defer()` or something equivalent. I never used this function so maybe it's not the right tool to do it, but on the principle I think each test file shouldn't influence the environment for the other tests

IndrajeetPatil:
> We should be careful with that because we define some variables in the initial testthat-file

@strengejacke Btw, this is no longer considered the best practice: https://github.com/r-lib/usethis/issues/1631

@etiennebacher Yeah, I think _withr_ is the way to go. I can give it a try later if you don't manage to get it to work.

etiennebacher:
@IndrajeetPatil I tried a few things but wrapping the code of a file in `withr::defer` is apparently not enough, and this case is not described in this [testthat vignette](https://testthat.r-lib.org/articles/test-fixtures.html#test-fixtures). Maybe I'll give it another try in the next days but I'm stuck for now

strengejacke:
> @strengejacke Btw, this is no longer considered the best practice: [r-lib/usethis#1631](https://github.com/r-lib/usethis/issues/1631)

But why?

etiennebacher:
> FYI I made a post on stackoverflow for this logistf/MCMCglmm bug

For info, this bug is due to the package `formula.tools`. A solution is detailed here: https://github.com/lrberge/fixest/issues/210

IndrajeetPatil:
> But why?

@strengejacke The [template ](https://github.com/r-lib/usethis/blob/main/inst/templates/testthat.R) file provides clues.

```r
# This file is part of the standard setup for testthat.
# It is recommended that you do not modify it.
#
# Where should you do additional test configuration?
# Learn more about the roles of various files in:
# * https://r-pkgs.org/tests.html
# * https://testthat.r-lib.org/reference/test_package.html#special-files
```

The setup for tests should be included in a separate `setup` file.

image

IndrajeetPatil:
This is hopeless. Every package seems to depend on some other package and thus can't be detached.

I think the easiest solution is going to be not include the test files that fail while computing code coverage. This can be done by adding the following tags at the start and end of each file:

```r
# nocov start
...
# nocov end
```

@etiennebacher I am closing this PR. Do you want to make a separate PR with the mentioned approach?
No worries if you can't. I can take care of it.

etiennebacher:
@IndrajeetPatil I think it's possible to specify some files to exclude from codecov in `covr::codecov()`. I'll make a few tries and open another PR if it works. Thanks for taking a look at it anyway

PR #580, opened 2022-06-08
fix clean_names() for mgcv::s() and gam::s()

### LINKED ISSUE #553: Question - Is there a way to find variables for smooth components in mgcv::gam?

etiennebacher:
Close #553
``` r
library(insight)
library(mgcv)
#> Loading required package: nlme
#> This is mgcv 1.8-40. For overview type 'help("mgcv-package")'.

clean_names('s(time)')
#> [1] "time"
clean_names('s(longitude, latitude)')
#> [1] "longitude, latitude"
clean_names('s(longitude, latitude, altitude)')
#> [1] "longitude, latitude, altitude"
clean_names('s(time, by = factor)')
#> [1] "time, factor"
clean_names('s(time, duration, by = factor)')
#> [1] "time, duration, factor"
clean_names('s(time, factor, bs = "fs")')
#> [1] "time, factor"
clean_names('s(factor, bs = "re")')
#> [1] "factor"
clean_names('s(factor, time, bs = "re")')
#> [1] "factor, time"

set.seed(2)
dat <- gamSim(1,n=400,dist="normal",scale=2)
#> Gu & Wahba 4 term additive model
d <- gam(y~x0+s(x1)+s(x2)+x3+s(x1,by = x2, k = -1), data=dat)
find_smooth(d, flatten = TRUE) |> clean_names()
#> [1] "x1"     "x2"     "x1, x2"
```

<sup>Created on 2022-06-08 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>


DominiqueMakowski:
what's the status of this?

strengejacke:
See discussion in https://github.com/easystats/insight/issues/553 - we haven't decided yet how to handle these issues. I have to revisit the discussion as well, but I'm not sure if this PR does on what we probably will decide.

etiennebacher:
This PR is useless as-is and it's not hard to redo if needed later.

PR #582, opened 2022-06-10
add codecov and exclude problematic tests

etiennebacher:
Follow up of #579, I test things in a PR because codecov can't run on my fork

### DIFF
--- a/.github/workflows/test-coverage.yml
+++ b/.github/workflows/test-coverage.yml
@@ -10,7 +10,7 @@ name: test-coverage
 
 jobs:
   test-coverage:
-    runs-on: macos-latest
+    runs-on: windows-latest
     env:
       GITHUB_PAT: ${{ secrets.GITHUB_TOKEN }}
 
--- a/.github/workflows/test-coverage.yml
+++ b/.github/workflows/test-coverage.yml
@@ -1,0 +1,31 @@
+# Workflow derived from https://github.com/r-lib/actions/tree/v2/examples
+# Need help debugging build failures? Start at https://github.com/r-lib/actions#where-to-find-help
+on:
+  push:
+    branches: [main, master]
+  pull_request:
+    branches: [main, master]
+
+name: test-coverage
+
+jobs:
+  test-coverage:
+    runs-on: macos-latest
+    env:
+      GITHUB_PAT: ${{ secrets.GITHUB_TOKEN }}
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - uses: r-lib/actions/setup-r@v2
+        with:
+          use-public-rspm: true
+
+      - uses: r-lib/actions/setup-r-dependencies@v2
+        with:
+          extra-packages: any::covr
+          needs: coverage
+
+      - name: Test coverage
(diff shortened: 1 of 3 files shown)


IndrajeetPatil:
I don't think you need to rely on _covr_ package.

Just add these tags to the start and end of problematic test files:

```
# nocov start
file content
# nocov end
```

tests/testthat/test-logistf.R
```diff
@@ -1,6 +1,7 @@
 if (requiet("testthat") &&
```
IndrajeetPatil, line 1:
```suggestion
# nocov start

if (requiet("testthat") &&
```

etiennebacher: aren't these comments used in the R files instead of the tests? do you have an example of this in other repos?

etiennebacher: https://github.com/r-lib/covr#exclusion-comments

etiennebacher:
> I don't think you need to rely on _covr_ package.

`covr` is installed anyway in the action so that shouldn't matter, right?

IndrajeetPatil:
See my reply [here](https://github.com/easystats/insight/pull/582#discussion_r894289312).

tests/testthat/test-logistf.R
etiennebacher: `testthat::skip_on_covr()` would work to skip the tests but what we need is to prevent `logistf` from being loaded when covr runs. This is why I initially put `!covr::in_covr`, the goal was to skip the whole file

etiennebacher: actually, putting `testthat::skip_on_covr()` at the top of the file seems to work

IndrajeetPatil:
I think you can skip a few more.

etiennebacher:
I don't know which ones. And `covr::codecov` works locally (except that I can't upload the results to codecov) but not on GHA

IndrajeetPatil:
Hmm, yeah, I can imagine. A lot of tests are being skipped depending on the system environment. 

You will probably need to comment out all of that and replace all `.runTest` with `TRUE`.

.github/workflows/test-coverage.yml
```diff
@@ -0,0 +1,31 @@
...
+  push:
+    branches: [main, master]
+  pull_request:
+    branches: [main, master]
+
+name: test-coverage
+
+jobs:
+  test-coverage:
+    runs-on: macos-latest
```
IndrajeetPatil, line 13: Can you try once with `runs-on: windows-latest`?

IndrajeetPatil: Daniel uses Windows machine, so whatever skipping logic is written, it is around that OS. So maybe there are better chances of seeing which tests fail.

Or are you also using Windows?

etiennebacher: I'm also on windows and it works, let's see if it works with windows-latest

IndrajeetPatil:
Merveilleux!

IndrajeetPatil (approved):

etiennebacher:
Great! Thanks for your help @IndrajeetPatil

Issue #585, opened 2022-07-02
`get_loglikelihood()` is slow because it calls `get_predicted()` with CIs


vincentarelbundock:
This is very slow, even for logit models with only 10K observations, and it slows down a bunch of things downstream in `performance` et al.

We could call `get_predicted(... ci = NULL)` to avoid computing CIs.

strengejacke:
Furthermore, profiled CIs are slow (from `model_parameters()`), that's why we have:

```r
  if (insight::n_obs(model) > 1e4 && identical(ci_method, "profile")) {
    message(insight::format_message("Profiled confidence intervals may take longer time to compute. Use 'ci_method=\"wald\"' for faster computation of CIs."))
  }
```

(https://github.com/easystats/parameters/blob/bd354f9f634e690bd8a51b0ebb39e17cc1a632a7/R/1_model_parameters.R#L623-L625)

strengejacke:
@easystats/maintainers Would you say it makes sense to set `ci = NULL` by default in `get_predicted()`?

bwiernik:
So long as we don't set it to NULL in modelbased, I'm fine with get_predicted() being NULL by default

DominiqueMakowski:
> So long as we don't set it to NULL in modelbased

I agree, insight by default should be fast and "minimal" and then we can leverage and re-purpose its functionalities downstream

vincentarelbundock:
Thanks all!

I didn't know about wald vs profile, so that is very useful info.

One thing to consider is that changing the default value of `ci` will probably break some people's code, and that's something we should probably avoid. We could just be more explicit when we use that function internally (which would be a good thing anyway).

strengejacke:
> One thing to consider is that changing the default value of `ci` will probably break some people's code, and that's something we should probably avoid. We could just be more explicit when we use that function internally (which would be a good thing anyway).

The function is rather new, and I'm not sure it's used that much outside _modelbased_ and _marginaleffects_. So _if_ we decide to change the default to `NULL`, it would be better now than later ;-)

vincentarelbundock:
Good point. Happy to defer to your preferences here.

PR #586, opened 2022-07-02
CIs are not needed

vincentarelbundock:
Since `modelsummary` switched to easystats, I've had several reports of GLM models be extremely slow (borderline unusable) when N>10K. I think part of the problem is the *IC statistics, which rely on this function, which calls the slow CI predictions.

This is a critical issue for me.

https://github.com/easystats/insight/issues/585

### DIFF
--- a/R/get_loglikelihood.R
+++ b/R/get_loglikelihood.R
@@ -179,7 +179,7 @@ get_loglikelihood.afex_aov <- function(x, ...) {
   w <- get_weights(x, null_as_ones = TRUE)
   dev <- stats::deviance(x)
   disp <- dev / sum(w)
-  predicted <- get_predicted(x, verbose = verbose)
+  predicted <- get_predicted(x, ci = NULL, verbose = verbose)
 
   # Make adjustment for binomial models with matrix as input
   if (info$is_binomial) {


bwiernik (approved):

IndrajeetPatil:
@vincentarelbundock Do you think we need to make a new release of _insight_ on CRAN?

I am not sure how crucial this change is for the release version of _modelsummary_.

strengejacke:
``` r
easystats::on_CRAN()
#> insight     7.2 weeks
#> datawizard  6.7 weeks
#> bayestestR  8.7 weeks
#> performance 1.7 weeks
#> parameters  4.9 weeks
#> effectsize  5.3 weeks
#> modelbased  4.7 weeks
#> correlation 6.2 weeks
#> see         1.7 weeks
#> report      18.6 weeks
```

<sup>Created on 2022-07-02 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

We could submit a new version of _insight_ before my holidays...

IndrajeetPatil:
Sounds good. I might also update _datawizard_ once the vignette PR is merged.

vincentarelbundock:
Cool, thanks all!

A new version might be very nice, as I'm gearing up to release v1.0.0 along with a JSS article.

I could do a bit more profiling before to see if there are other low hanging fruits. When would you need to move ahead with this @strengejacke ?

strengejacke:
I'm away from next Friday, so either we submit until next Thursday, or by the end of July.

vincentarelbundock:
If you think a release now could be worth it and isn't too much trouble, that's cool. But no pressure at all. I need to take some time off anyway, so I won't be able to do anymore for a while.

Also, the `predict=NULL` default would likely require adjusting the test suite, so not sure anyone is going to jump on that right away.

strengejacke:
_modelbased_ explicitely sets CIs, performance doesn't require them and _bayestestR_ only provides methods. So I think it should be ok to set CI to NULL, will do this and run tests, and then let's see what rev-dep checks will say. And from _marginaleffects_ side, it's also OK to set `ci = NULL`?

vincentarelbundock:
Hmmm, not 100% sure about `marginaleffects`, and I won't have a computer to check for a few days. But don't worry about it. I'll look into that later.

PR #591, opened 2022-07-15
scale message

vincentarelbundock:
`get_predicted()` raises an informative message when there is a matrix column in the terms. However, sometimes we supply our own `data`, and the message is still raised even if that `data` does *not* include a matrix column.

### DIFF
--- a/R/get_predicted_args.R
+++ b/R/get_predicted_args.R
@@ -56,22 +56,25 @@
 
 
   # Intermediate step, check data classes  ------
+  # if data=NULL, check terms
+  # if data!=NULL, check data
   ############################################################################
 
-  tryCatch(
-    {
-      data_classes <- attributes(stats::terms(x))$dataClasses
-      if (any(grepl("matrix", data_classes, fixed = TRUE)) && isTRUE(verbose)) {
-        message(format_message(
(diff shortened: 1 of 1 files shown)


R/get_predicted_args.R
```diff
@@ -56,22 +56,25 @@
...
-
-    }
-  )
+  if (is.null(data)) {
+    flag_matrix <- tryCatch(
+      any(grepl("matrix", attributes(stats::terms(x))$dataClasses, fixed = TRUE)),
+      error = function(e) FALSE)
+  } else {
+    flag_matrix <- tryCatch(
+      any(sapply(data, function(x) "matrix" %in% class(x))),
```
bwiernik, line 69:
```suggestion
      any(sapply(data, function(x) inherits(x, "matrix"))),
```

bwiernik (changes requested):

bwiernik:
One minor comment then looks good to me. You can ignore the windows test errors for the moment

vincentarelbundock:
Good call on `inherits`! I made the change.

I think the repo now "protects" the main branch, so I won't be able to merge until there's an "official" review (not sure exactly how this works).

IndrajeetPatil:
Thanks, @vincentarelbundock !

PR #595, opened 2022-07-24
`find_formula` evaluates symbol in call

vincentarelbundock:
Problem: sometimes users assign a formula to an object, and then feeds that object to the formula parameter of a fitting function. In the `selection` model type, this poses problems because we extract the formula from the model call, and the apply `stats::as.formula` directly to that. But if the call includes a symbol rather than the formula itself, this breaks. We need to evaluate the symbol before calling `as.formula`.

This is a simple PR with a `tryCatch` and a 2nd attempt with `eval()`

### DIFF
--- a/R/find_formula.R
+++ b/R/find_formula.R
@@ -316,9 +316,19 @@ find_formula.marginaleffects <- function(x, verbose = TRUE, ...) {
 #' @export
 find_formula.selection <- function(x, verbose = TRUE, ...) {
   model_call <- parse(text = deparse(get_call(x)))[[1]]
+  # formulas directly in the call
+  f_selection <- tryCatch(stats::as.formula(model_call$selection), error = function(e) NULL)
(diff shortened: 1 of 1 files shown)


bwiernik:
@vincentarelbundock my comment didn't post for some reason 

```suggestion

  # first, assume that formulas are specified directly in the call
  # if that errors, assume that formulas are symbols (assigned to an object, which is then used in the call)
  f_selection <- tryCatch(
    stats::as.formula(model_call$selection), 
    error = function(e) stats::as.formula(eval(model_call$selection))
  )
  f_outcome <- tryCatch(
    stats::as.formula(model_call$outcome), 
    error = function(e) stats::as.formula(eval(model_call$outcome))
  )
```

We can skip the intermediate NULL checking

vincentarelbundock:
Good call, thanks!

I just too used to making sure the error case is always "valid", but returning a different error here is fine.

PR #609, opened 2022-08-11
format_message accepts tokens

strengejacke:
This PR adds minor formatting options to `format_message()`:

- `{.b boldtext}`
- `{.i emphasizetext}`
- `{.url url.toweb.site}`
- `{.pkg easystats}`

Example:

```r
library(insight)
x <- "This is a very long string that allow {.b bold} formatting"
cat(format_message(x))
```

Limitations: it's hard to detect the "real" (without format tags) streng length for line breaks. Thus, it can happen that lines are wrapped at an earlier length than expected. Furthermore, if you have multiple words in a format tag (`{.b one two three}`), a line break might occur in this tag, and the formatting no longer works.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Type: Package
 Package: insight
 Title: Easy Access to Model Information for Various Model Objects
-Version: 0.18.2
+Version: 0.18.2.1
 Authors@R: 
     c(person(given = "Daniel",
              family = "Lüdecke",
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,4 +1,11 @@
-# insight 0.18.1
+# insight 0.18.3
+
+## Changes to functions
+
+* `format_message()` gets some additional formatting features. See 'Details'
+  in `?format_message` for more information and some current limitations.
+
+# insight 0.18.2
 
 ## New functions
 
--- a/R/format_message.R
+++ b/R/format_message.R
@@ -52,23 +52,30 @@ format_message <- function(string, ..., line_length = 0.9 * options()$width) {
   lsub <- 0
   tmp_string <- string
 
-  # check if string contains any formatting token
+  # supported tokens, create regex pattern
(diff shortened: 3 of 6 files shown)


strengejacke:
@IndrajeetPatil You may check out this PR and test:

```r
library(insight)
x <- "This is a very long string that allow {.b bold} formatting, {.i emphasize} and of course fancy {.pkg packagenames} or even text styles as URLs, like {.url easystats.github.io}. You like?"
cat(format_message(x))
```

IndrajeetPatil:
This is really cool, Daniel! 😍

Screenshot 2022-08-12 at 07 54 09

strengejacke:
I'm not sure what common used tags would be, I thought these are nice to have. We could of course add all color tags, but I think bold/italic is ok for emphasizing.

IndrajeetPatil:
I think the most important ones (for preparing user-facing messages) would be `.fn` (function) and `.code` (code).

strengejacke:
yeah, I just that that typing {.code mean} is more work than typing 'mean()' - thus, didn't see much benefit for having those tags, too. but it's no big deal to add them.

IndrajeetPatil:
In a message, the _code_ should be easily visually distinguishable from the rest of the _text_, even if we, as developers, need to type more to make them so.

strengejacke:
> message, the _code_ should be easily visually distinguishable from the rest of the _text_, even if we, as developers, need to type more to make them so.

Yes, I agree, but would you type `{.fn mean}` (10 chars) instead of `'mean()'` (8 chars)? Or `{.code x=5}` (11 chars) vs. `'x=5'` (5 chars)?

IndrajeetPatil:
Hmm, I just checked with {dplyr} and it doesn't use monospace font for printing code or anything. So I think we just need to wrap the code in backticks. Not necessary to create tokens for these then. :)

Screenshot 2022-08-12 at 08 23 03

strengejacke:
If we could also change the font family in the output, than a tag would indeed make much more sense. But in either case, we can add more tokens later, if we want to.

Issue #612, opened 2022-08-14
`get_modelmatrix()` breaks when `data` includes only one factor level


vincentarelbundock:
`get_modelmatrix()` breaks when the `data` argument does not include at least two factor levels:

``` r
library(insight)

dat <- mtcars
dat$cyl <- factor(dat$cyl)
mod <- glm(vs ~ cyl, data = dat, family = binomial)

# works
get_modelmatrix(mod, data = head(dat, 3))
#>               (Intercept) cyl6
#> Mazda RX4               1    1
#> Mazda RX4 Wag           1    1
#> Datsun 710              1    0
#> attr(,"assign")
#> [1] 0 1
#> attr(,"contrasts")
#> attr(,"contrasts")$cyl
#> [1] "contr.treatment"

# breaks
get_modelmatrix(mod, data = head(dat, 1))
#> Error in `contrasts<-`(`*tmp*`, value = contr.funs[1 + isOF[nn]]): contrasts can be applied only to factors with 2 or more levels
```

This is of course because `stats::model.matrix()` does not support this either:

``` r
model.matrix(mod, data = head(dat, 1))
#> Error in `contrasts<-`(`*tmp*`, value = contr.funs[1 + isOF[nn]]): contrasts can be applied only to factors with 2 or more levels
```

Would it make sense for `get_modelmatrix()` to include some guardrails, and perhaps pad/unpad the data with missing factors when necessary?

First reported here: <https://github.com/vincentarelbundock/marginaleffects/issues/455>

strengejacke:
Possible duplicate of #316

vincentarelbundock:
hahaha

so, shall we pad?

strengejacke:
(and I think we haven't even touched [contrasts](https://github.com/strengejacke/ggeffects/blob/fef45e768f8345bae81137db6d583cf038221048/R/vcov.R#L264))

bwiernik:
I suggest we pad the model matrix with rows containing all factor levels from the original data, make the matrix, then remove those rows. That would be an awesome improvement on model.matrix() functionality

vincentarelbundock:
I can give this a shot eventually.

strengejacke:
@vincentarelbundock should we ass the pad/unpad stuff to other `get_modelmatrix()` methods, too?

strengejacke:
edit, "add", not "ass" 😬

vincentarelbundock:
[...]

PR #613, opened 2022-08-15
pad factors in get_modelmatrix()

vincentarelbundock:
PR for Issue #612 

The test suite segfaults on my computer (certainly my fault), so it is extra important to way for all the CI tests to show green.


vincentarelbundock:
Tests pass, but this doesn't quite work yet. Some weird environment-related problems, especially when called within functions or from downstream packages. Investigating.

vincentarelbundock:
turns out things were fine.

merged without review because the PR is extra-simple, easy to revert and extend if someone finds an issue.

PR #616, opened 2022-08-23
get_statistic.coxph returns wrong values

vincentarelbundock:
Problem reported here: https://github.com/vincentarelbundock/modelsummary/issues/547

Might need a fix in `parameters` as well.


vincentarelbundock:
No need for a `parameters` update after all, this is enough.

I also added another commit to make the `get_statistics.coxph` output similar to the default method (with attributes, etc.)

PR #627, opened 2022-08-31
fix #626

### LINKED ISSUE #626: `estimate_expectation()` fails for logistic regression with factor response

strengejacke:


IndrajeetPatil (commented):
Do we need to soft import the entire package _carData_ for a single dataset? 

Can we not just include it only in this test?

strengejacke:
> Do we need to soft import the entire package _carData_ for a single dataset?
> 
> Can we not just include it only in this test?

How?

IndrajeetPatil:
@strengejacke Just do-

```r
dput(carData::Mroz)
```

And copy-paste the output to test file.

That said, not sure how big this dataset is.

strengejacke:
ah, you mean an internal copy or create that data inside the script.... maybe for test purposes, we can reduce that data set

vincentarelbundock:
FWIW, I use `read.csv()` to RDatasets in many tests. It requires internet connectivity, of course, but I've never run into issues. `carData` is there:

https://vincentarelbundock.github.io/Rdatasets/

PR #637, opened 2022-09-18
prepare CRAN release

strengejacke:


strengejacke:
on its way to CRAN...

PR #648, opened 2022-09-25
add normal and wald types

strengejacke:
Usually, `parameters::degrees_of_freedom()` and `insight::get_df()` are expected to do the same thing. However, currently the design of those two functions differ. `parameters::degrees_of_freedom()` offers more options to extract df's. This PR aims at bringing `insight::get_df()` on par with `parameters::degrees_of_freedom()`, so the latter can be fully replaced by `insight::get_df()`. I think it doesn't make sense and is rather confusing to have two methods which do different things for certain models.

@easystats/core-team


R/get_df.R
```diff
@@ -5,11 +5,21 @@
...
-#' model-based degrees of freedom, i.e. the number of (estimated) parameters.
+#' @param type Can be `"residual"`, `"wald"`, `"normal"`, `"analytical"`, or
+#'   `"model"`.
+#'
+#' - `"residual"` tries to extract residual degrees of freedoms. If residual
+#'   degrees of freedom could not be extracted, returns analytical degrees of
+#'   freedom, i.e. `n-k` (number of observations minus number of parameters).
+#' - `"wald"` for models with z-statistic, returns `"Inf"`. Else, tries to
+#'   extract residual degrees of freedoms. If residual degrees of freedom could
+#'   not be extracted, returns `"Inf"`.
```
bwiernik, line 16:
```suggestion
#'   extract residual degrees of freedoms. If residual degrees of freedom
#'   cannot be extracted, returns `"Inf"`.
```

R/get_df.R
```diff
@@ -5,11 +5,21 @@
...
+#' - `"analytical"` returns analytical degrees of freedom, i.e. `n-k`
```
bwiernik, line 17: What's the difference between residual and analytical? They sound the same to me from this description

bwiernik (changes requested):

R/get_df.R
strengejacke: analytical is always `n_obs() - n_param()` (n-k). not sure if which cases this may differ from residual df?

strengejacke:
@bwiernik @mattansb what should "analytical" df return for models with z-statistic? Still `n-k`, or `Inf`? If the latter, `"analytical"` would be almost the same `"wald"`. See docs:

```
- `"residual"` tries to extract residual degrees of freedoms. If residual
   degrees of freedom cannot be extracted, returns analytical degrees of
   freedom, i.e. `n-k` (number of observations minus number of parameters).
 - `"wald"` for models with z-statistic, returns `"Inf"`. Else, tries to
   extract residual degrees of freedoms. If residual degrees of freedom
   cannot be extracted, returns `"Inf"`.
 - `"analytical"` for models with z-statistic, returns `"Inf"`. Else, returns
   analytical degrees of freedom, i.e. `n-k` (number of observations minus
   number of parameters).
 - `"normal"` always returns `"Inf"`.
 - `"model"` returns model-based degrees of freedom, i.e. the number of
   (estimated) parameters.
```

bwiernik:
I think the only issue is that residual is poorly described. It's whatever the model returns as the df, whereas analytical is always exactly n - params.

strengejacke:
Should we be more strict/consistent here?

strengejacke:
I think analytical and residual df are actually the same, so maybe we should simplify this.

Then we would have
- "normal" returns Inf.
- "wald" returns residual df for models with t-statistic, and Inf for all other models
- "analytical" (aka "residual") returns n-k (number of observations minus number of parameters) and `Inf` if residual degrees of freedom could not be extracted

strengejacke:
@vincentarelbundock 
would you say this model has a t- or z-statistic? Summary says t, but it has Poisson link:

``` r
library("fixest")
data(trade)
m1 <- femlm(Euros ~ log(dist_km) | Origin + Destination + Product, data = trade)

summary(m1)
#> ML estimation, family = Poisson, Dep. Var.: Euros
#> Observations: 38,325 
#> Fixed-effects: Origin: 15,  Destination: 15,  Product: 20
#> Standard-errors: Clustered (Origin) 
#>              Estimate Std. Error  t value  Pr(>|t|)    
#> log(dist_km) -1.52775   0.115627 -13.2127 < 2.2e-16 ***
#> ---
#> Signif. codes:  0 '***' 0.001 '**' 0.01 '*' 0.05 '.' 0.1 ' ' 1
#> Log-Likelihood: -7.133e+11   Adj. Pseudo R2: 0.760389
#>            BIC:  1.427e+12     Squared Cor.: 0.60377

m1$family
#> [1] "poisson"
```

<sup>Created on 2022-09-26 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

mattansb:
For some models it depends how you define "number of parameters" - thinking of mixed models. If I have 1000 obs from 10 subjects and `y ~ 1 + time + (1 + time | Subject)` `lme4` has 6 parameters (2 fixed, 2 variance, 1 random covariance, 1 sigma), but arguably there are more - the 2 * 10 individual random parameters. 

Something to think about?

PR #652, opened 2022-09-29
Improve performance of `object_has_rownames()` 

etiennebacher:
``` r
# current 

object_has_rownames <- function(x) {
  if (!is.data.frame(x)) {
    stop("Only dataframe objects are allowed.", call. = FALSE)
  }
  
  !identical(as.character(seq_len(nrow(x))), rownames(x))
}

# new

new_object_has_rownames <- function(x) {
  if (!is.data.frame(x)) {
    stop("Only dataframe objects are allowed.", call. = FALSE)
  }
  
  !all(is.numeric(attributes(x)$row.names))
}


out <- list()
for (i in 1:100000) {
  out[[i]] <- iris
}

out <- data.table::rbindlist(out)
nrow(out)
#> [1] 15000000

bench::mark(
  old = object_has_rownames(out),
  new = new_object_has_rownames(out)
)
#> # A tibble: 2 × 6
#>   expression      min   median   `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr> <bch:tm> <bch:tm>       <dbl> <bch:byt>    <dbl>
#> 1 old           13.5s    13.5s      0.0739     484MB   0.0739
#> 2 new             1µs    1.6µs 359776.            0B   0
```

<sup>Created on 2022-09-29 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

### DIFF
--- a/R/utils_object.R
+++ b/R/utils_object.R
@@ -84,5 +84,5 @@ object_has_rownames <- function(x) {
     stop("Only dataframe objects are allowed.", call. = FALSE)
   }
 
-  !identical(as.character(seq_len(nrow(x))), rownames(x))
+  !all(is.numeric(attributes(x)$row.names))
 }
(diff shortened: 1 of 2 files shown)


IndrajeetPatil (approved):
Wow, that's orders of magnitude improvement! Thanks.

bwiernik:
Hmmm, I'm not sure this is strict enough though. For example, `psych::bfi` has numeric rownames that are participant IDs. How much performance cost is there to also check if the first element is `1` or the first few are `1:6`?

etiennebacher:
Right, I'll make a new PR to take this into account

PR #653, opened 2022-09-30
Fix `object_has_rownames()`

etiennebacher:
Make `object_has_rownames()` stricter (but still more performant), cf #652 :
``` r
# current 

object_has_rownames <- function(x) {
  if (!is.data.frame(x)) {
    stop("Only dataframe objects are allowed.", call. = FALSE)
  }
  
  !identical(as.character(seq_len(nrow(x))), rownames(x))
}

# new

new_object_has_rownames <- function(x) {
  if (!is.data.frame(x)) {
    stop("Only dataframe objects are allowed.", call. = FALSE)
  }
  
  !identical(attributes(x)$row.names, 1:nrow(x))
}


out <- list()
for (i in 1:100000) {
  out[[i]] <- iris
}

out <- data.table::rbindlist(out)
nrow(out)
#> [1] 15000000

bench::mark(
  old = object_has_rownames(out),
  new = new_object_has_rownames(out)
)
#> # A tibble: 2 × 6
#>   expression      min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr> <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 old           12.3s    12.3s    0.0814     484MB   0.0814
#> 2 new          44.8ms   53.2ms    6.77       114MB   1.35
```

<sup>Created on 2022-09-30 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

Is it ok for you @bwiernik? I added a test for `psych::bfi`

### DIFF
--- a/R/utils_object.R
+++ b/R/utils_object.R
@@ -84,5 +84,5 @@ object_has_rownames <- function(x) {
     stop("Only dataframe objects are allowed.", call. = FALSE)
   }
 
-  !all(is.numeric(attributes(x)$row.names))
+  !identical(attributes(x)$row.names, 1:nrow(x))
 }
--- a/R/utils_object.R
+++ b/R/utils_object.R
@@ -84,5 +84,5 @@ object_has_rownames <- function(x) {
     stop("Only dataframe objects are allowed.", call. = FALSE)
   }
 
-  !identical(attributes(x)$row.names, 1:nrow(x))
+  !identical(attributes(x)$row.names, seq_len(nrow(x)))
 }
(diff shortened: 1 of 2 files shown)


R/utils_object.R
```diff
@@ -84,5 +84,5 @@ object_has_rownames <- function(x) {
     stop("Only dataframe objects are allowed.", call. = FALSE)
   }
 
-  !all(is.numeric(attributes(x)$row.names))
+  !identical(attributes(x)$row.names, 1:nrow(x))
```
IndrajeetPatil, line 87:
```suggestion
  !identical(attributes(x)$row.names, seq_along(x))
```

``` r
1:nrow(data.frame())
#> [1] 1 0

seq_along(data.frame())
#> integer(0)
```

<sup>Created on 2022-09-30 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

etiennebacher: I used `seq_len(nrow(x))` instead, like the original code (doesn't change perf)

IndrajeetPatil (approved):
Thanks!

Issue #645, opened 2022-09-21
`get_data.brmsfit` does not extract all columns when `nl=TRUE`


vincentarelbundock:
`insight::get_data.brmsfit` does not extract all variables when `bf()` includes the `nl=TRUE` option and some terms in parentheses in “auxiliary” equations. In this example, we are missing the `gear` variable.

I don’t know `brmsfit` objects very well, so I’m not exactly sure how to diagnose this problem, but my guess is that the problem arises in `find_predictors`, and I see a note which may be relevant here:

<https://github.com/easystats/insight/blob/main/R/find_predictors.R#L238>

``` r
library(brms)
library(insight)

mod <- brm(
    data = mtcars,
    backend = "cmdstanr",
    family = cumulative(probit),
    formula = bf(
        cyl ~ 1 + mpg + drat + gearnl,
        gearnl ~ 0 + (1 | gear),
        nl = TRUE))

insight::get_data(mod) |> head()
#>                   cyl  mpg drat
#> Mazda RX4           6 21.0 3.90
#> Mazda RX4 Wag       6 21.0 3.90
#> Datsun 710          4 22.8 3.85
#> Hornet 4 Drive      6 21.4 3.08
#> Hornet Sportabout   8 18.7 3.15
#> Valiant             6 18.1 2.76

insight::find_predictors(mod)
#> $conditional
#> [1] "mpg"  "drat"
```

vincentarelbundock:
https://github.com/easystats/insight/pull/661

PR #655, opened 2022-10-02
Issue #645

vincentarelbundock:
In Issue #645 I noted that when users fit a `brms` model with `nl=TRUE` and multiple equations, the `get_data` function does not extract all predictors because `find_predictors()` does not identify them all.

This is a hack to fix this. I say "hack" because all the variables in the non-linear equations get inserted in the "conditional" vector of the `get_predictors()` output list.

It is not clear to me where these variables should go anyway, and the `get_predictors.default()` method assumes a single formula. So this might be the best we can do without creating a complex `find_predictors.brmsfit()` from scratch (I don't have the bandwith for that now).

At least this allows us to identify all the predictors and extract them via `get_data.brmsfit()`.

Would be curious to know if people have thoughts before I merge.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -165,7 +165,7 @@ VignetteBuilder:
     knitr
 Encoding: UTF-8
 Language: en-US
-RoxygenNote: 7.2.1
+RoxygenNote: 7.2.1.9000
 Roxygen: list(markdown = TRUE)
 Config/testthat/edition: 3
 Config/Needs/website:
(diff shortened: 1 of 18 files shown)


bwiernik:
That's fine for now I think. Can you put a TODO in the code to make something better and modify the issue on GitHub to clarify what we should do instead?

I'm not sure what's the best structure to use instead. Maybe a slot `nl` or multiple slots named by their response variables? Or maybe those two combined--a `nl` slot with sub slots for each formula?

strengejacke:
See `find_parameters.nlmerMod`, we have an element `"nonlinear"`. That might be better, and probably would prevent the tests from failing.

Issue #658, opened 2022-10-05
`find_formula` fails when applied to `lme` inside `lapply`


iago-pssjd:
I come here from https://github.com/vincentarelbundock/modelsummary/issues/559.
I adapt the reprex
```r
library(insight)
library(nlme)
models <-lapply(c("", " + Sex"), \(.x) lme(as.formula(paste0("distance  ~ age", .x)), random = ~ 1, data = Orthodont))
lapply(models, find_formula)
Error in paste0("distance  ~ age", .x) : object '.x' not found
```
I would say that this seems to be a duplicate of issue #309, but next line works
```r
find_formula(lme(as.formula(paste0("distance  ~ age", "")), random = ~ 1, data = Orthodont))
$conditional
distance ~ age
<environment: 0x55c02e144f78>

$random
~1
<environment: 0x55c02e144f78>

attr(,"class")
[1] "insight_formula" "list"         
```

Thanks!

iago-pssjd:
Maybe solvable changing `eval(x$call$fixed) ` in
https://github.com/easystats/insight/blob/6fdc6016313f49caeeba5feb2643be6ad2623a1d/R/find_formula.R#L1198-L1200
with `formula(x$terms)`?

vincentarelbundock:
> @strengejacke I don't see your point. I only modify the fixed effects part of `find_formula.lme`

One question is: if the random part is something like `random=~x1+x2`, will your change properly classify those as "random", or will it stuff the "x1" and "x2" variables in the "fixed" part by mistake?

strengejacke:
Ah ok, didn't test with the PR.

iago-pssjd:
@vincentarelbundock 
Let's see with the example:
```r
library(insight)
library(nlme)
models <-lapply(c("", " + Sex"), \(.x) lme(as.formula(paste0("distance  ~ age", .x)), random = ~ Sex, data = Orthodont))
find_formula(models[[1]])
$conditional
distance ~ age
<environment: 0x55a163099e58>

$random
~Sex
<environment: 0x55a163aea038>

attr(,"class")
[1] "insight_formula" "list"      
```
In this case, it seems to work. I may try some more complex example.

vincentarelbundock:
Great!

But I bet the code breaks if the lapply modifies the random component instead of the fixed.

Is that a big deal?

iago-pssjd:
@vincentarelbundock 
> if the lapply modifies the random component 
[...]

PR #659, opened 2022-10-06
fix issue 658 by reproducing formula function

iago-pssjd:
Close issue #658. The idea, as told there, is to do `fm <- formula(x$terms)` instead of `fm <- eval(x$call$fixed) `, but this would need to include `importFrom("stats", formula)` in the NAMESPACE, and since there is no `importFrom` there, I assumed you prefer not use them. Therefore, instead of using `formula(....)`, I wrote the  code of `stats:::formula.terms` there.

### DIFF
--- a/R/find_formula.R
+++ b/R/find_formula.R
@@ -1196,7 +1196,14 @@ find_formula.sem <- function(x, verbose = TRUE, ...) {
 
 #' @export
 find_formula.lme <- function(x, verbose = TRUE, ...) {
-  fm <- eval(x$call$fixed)
+  env <- environment(x$terms)
+  fm <- x$terms
+  attributes(fm) <- list(class = "formula")
+  environment(fm) <- if(is.null(env)) {
+	  globalenv()
+  } else {
+	  env
+  }
   fmr <- eval(x$call$random)
   ## TODO this is an intermediate fix to return the correlation variables from lme-objects
   fcorr <- x$call$correlation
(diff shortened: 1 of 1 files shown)


vincentarelbundock:
Thanks for opening a pull request!

You don't need to use `importFrom` or to re-write the code yourself. Instead, you can simply call the function from its namespace:

```r
fm <- stats::formula(x$terms)
```

Are we sure that the two versions will **always** give the same results? What if there are variables in both the random and fixed components/formulas? Do both codes give the same results in a model estimated outside a `lapply`?

I'm not sure what the original intent of the code was in that respect, so it would be important to make sure nothing changes in different models.

iago-pssjd:
@vincentarelbundock thanks for the point of using `stats::formula`. I believe this is more a stylistic which you, the package developers, know better than me your preferences.

> Are we sure that the two versions will always give the same results? What if there are variables in both the random and fixed components/formulas? Do both codes give the same results in a model estimated outside a lapply?
> 
> I'm not sure what the original intent of the code was in that respect, so it would be important to make sure nothing changes in different models.

It seems so, at least with the original formulas. I just have built, installed and checked `--as-cran` the package without any complaints. It lacks a revdep check, so it does not changes other packages behaviours, but I am not sure how and do not have enough time to do it.

vincentarelbundock:
Thanks a lot for the PR!

I merged it and added a test here: https://github.com/easystats/insight/pull/660

PR #661, opened 2022-10-07
Issue #645 v2

vincentarelbundock:
I broke the merge on my previous PR, so this is a new attempt.


bwiernik:
Gotta love git

vincentarelbundock:
> Gotta love git

What's "nice" about git is I'm pretty sure there was an objectively correct way to do this. So I get to feel both (1) annoyed by the extra work and (2) dumb for not mastering the tool.

Also, I can't figure out why the tests fail on older platforms. Failures seem completely unrelated to this PR

vincentarelbundock:
Any idea why this super minimal PR would raise a new error?

```r
── Error (test-get_predicted-clm.R:6:3): get_predicted.default - ordinal - match CI ──
Error in `.Call("get_fitted", eta1, eta2, link, lambda)`: "get_fitted" not resolved from current namespace (ordinal)
Backtrace:
    ▆
 1. └─ordinal::clm(rating ~ temp * contact, data = wine) at test-get_predicted-clm.R:6:2
 2.   └─ordinal:::clm.fit.default(clm.struct)
 3.     └─ordinal:::clm.fit.NR(rho, y$control)
 4.       └─rho$clm.nll(rho)
 5.         └─ordinal:::getFittedC(rho$eta1, rho$eta2, rho$link, rho$par[length(rho$par)])
```

tagging @strengejacke and (testing maestro) @IndrajeetPatil

IndrajeetPatil:
I can have a closer look later when I have my computer on me, but just looking at this error, it feels like the problem is with _ordinal_ package itself? 

Error in `.Call("get_fitted", eta1, eta2, link, lambda)`: "get_fitted" not resolved from current namespace (ordinal)

IndrajeetPatil:
Like R can't seem to figure out which C routine to call.

IndrajeetPatil:
@vincentarelbundock It's the same issue as this one: https://stackoverflow.com/questions/18192225/r-error-message-package-error-functionname-not-resolved-from-current-nam

The method being called is defined [here](https://github.com/runehaubo/ordinal/blob/6b279bfc7aed755bf4647de3cd6faf4da1ef282e/src/get_fitted.c#L30).

But looking at it and looking at the trace, I don't see anything that is wrong with _ordinal_.

``` r
ordinal:::get_fitted
#> $name
#> [1] "get_fitted"
#> 
#> $address
#> <pointer: 0x600002faa560>
#> attr(,"class")
#> [1] "RegisteredNativeSymbol"
#> 
#> $dll
#> DLL name: ordinal
#> Filename:
#>         /Library/Frameworks/R.framework/Versions/4.2-arm64/Resources/library/ordinal/libs/ordinal.so
#> Dynamic lookup: FALSE
#> 
#> $numParameters
#> [1] 4
#> 
#> attr(,"class")
#> [1] "CallRoutine"      "NativeSymbolInfo"
```

<sup>Created on 2022-10-10 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

Sorry, I can't be more helpful; don't have much experience working with compiled code in R packages.

strengejacke:
> Any idea why this super minimal PR would raise a new error?

Looks like this only happens on certain machines, so could we skip these tests on the affected OSses?

strengejacke (approved):

bwiernik:
Thanks!

vincentarelbundock:
Thanks!

strengejacke:
Thanks!

PR #671, opened 2022-10-17
Fixes easystats/performance#432 (missing outlier plot when scaling data)

rempsyc:
Fixes easystats/performance#432 (missing outlier plot when scaling data)

``` r
# Scale data
library(dplyr)
mtcars2 <- mtcars %>%
  mutate(across(everything(), scale))

# Create model
m2 <- lm(mpg ~ wt + cyl + gear + disp, data = mtcars2)

# Verify that it fails to plot outliers with current version
performance::check_model(m2)
```

``` r
# Load library
devtools::load_all(".../forks/insight")

# Verify that it works with dev version
performance::check_model(m2)
```

<sup>Created on 2022-10-16 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

### DIFF
--- a/R/get_predicted.R
+++ b/R/get_predicted.R
@@ -319,7 +319,23 @@ get_predicted.lm <- function(x,
   args <- .get_predicted_args(x, data = data, predict = predict, verbose = verbose, ...)
 
   predict_function <- function(x, data, ...) {
-    stats::predict(x, newdata = data, interval = "none", type = args$type, se.fit = FALSE, ...)
+    stats::predict(x, newdata = data, interval = "none",
+                   type = args$type, se.fit = FALSE, ...)
+  }
+
+  # 0. step: convert matrix variable types attributes to numeric, if necessary
+  dataClasses <- attributes(x[["terms"]])$dataClasses
+  if ("nmatrix.1" %in% dataClasses) {
+    length.dataClasses <- length(dataClasses)
+    names.dataClasses <- names(dataClasses)
+    attributes(x[["terms"]])$dataClasses <- setNames(
+      unlist(lapply(dataClasses, function(x) {
+        if (x == "nmatrix.1") {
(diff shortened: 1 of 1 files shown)


R/get_predicted.R
```diff
@@ -319,7 +319,23 @@ get_predicted.lm <- function(x,
...
+    length.dataClasses <- length(dataClasses)
+    names.dataClasses <- names(dataClasses)
+    attributes(x[["terms"]])$dataClasses <- stats::setNames(
+      unlist(lapply(dataClasses, function(x) {
+        if (x == "nmatrix.1") {
+          "numeric"
+        } else {
+          x
+        }
+      })), names.dataClasses)
```
IndrajeetPatil, line 338:
```suggestion
    attributes(x[["terms"]])$dataClasses <- stats::setNames(
      unlist(lapply(dataClasses, function(x) if (x == "nmatrix.1") "numeric" else x)), 
      names.dataClasses
    )
```

IndrajeetPatil (changes requested):
Thanks. LGTM!

Do you mind adding a unit test as well to insure we won't reintroduce this bug?

R/get_predicted.R
rempsyc: This does put the line at 80 characters, so a bit borderline for lintr but still OK I guess 😬

IndrajeetPatil: We don't use the 80 character limit, but rather 120.

IndrajeetPatil: You can see our lint config here: https://github.com/easystats/workflows/blob/main/.github/workflows/lint.yaml

rempsyc:
Tests added! :D I checked for three things, let me know if that's enough or we need more.

IndrajeetPatil:
Thanks!

PR #675, opened 2022-10-23
prepare CRAN submission

strengejacke:


IndrajeetPatil:
The CRAN submission file should be added to .gitignore, btw.

PR #683, opened 2022-11-17
Implement `has_one_value()`

etiennebacher:
Coming from https://github.com/easystats/datawizard/pull/318


strengejacke:
Just one thing... Should we have an `na.rm` argument, just in case someone wants to test only for non-missing values?

etiennebacher:
I'm not sure, wouldn't it be better to explicitly remove the NA from the vector you want to use in `has_one_value()`?

NAMESPACE
```diff
@@ -1564,6 +1564,7 @@ export(get_variance_residual)
 export(get_variance_slope)
 export(get_weights)
 export(has_intercept)
+export(has_one_value)
```
IndrajeetPatil, line 1567: Maybe `has_single_value()` is a better name? WDYT?

etiennebacher: Agreed (but can't modify now, I will change it later)

Issue #689, opened 2022-11-25
`get_data()` produces incorrect results with transformations


vincentarelbundock:
When a model formula includes transformations, `get_data()` returns incorrect results:

``` r
library(insight)

x <- abs(rnorm(100, sd=5)) + 5
y <- exp(2 + 0.3 * x + rnorm(100, sd=0.4))
dat <- data.frame(y, x)
mod <- lm(log(y) ~ log(x), data = dat)

head(insight::get_data(mod))
              y             x
    1  62.01648  0.000000e+00
    2  21.08598  8.881784e-16
    3  41.99490  8.881784e-16
    4  28.51719  8.881784e-16
    5 289.96846  1.776357e-15
    6 120.86293 -1.776357e-15

head(dat)
              y         x
    1  69.07015  7.053672
    2  26.25120  5.165226
    3  49.20110  7.206200
    4  34.23304  5.715846
    5 301.20648 11.238027
    6 129.50084  8.637918
```

I know recovering from environment is not always ideal, but should we change `get_data()` to always try to do that first before falling back on `model.frame`? This would ensure that the actual original data gets returned in most cases.

bwiernik:
What does `model.frame()` return here?

vincentarelbundock:
`model.frame()` returns the logged version, which isn't what we want (I think).

``` r
library(insight)

set.seed(1024)

x <- abs(rnorm(100, sd=5)) + 5
y <- exp(2 + 0.3 * x + rnorm(100, sd=0.4))
dat <- data.frame(y, x)
mod <- lm(log(y) ~ log(x), data = dat)

head(insight::get_data(mod), 2)
             y x
    1 89.32396 0
    2 42.38241 0

head(dat, 2)
             y        x
    1 98.21727 8.893314
    2 49.32979 6.947382

head(model.frame(mod))
        log(y)   log(x)
    1 4.587182 2.185300
    2 3.898528 1.938365
    3 6.348238 2.719253
    4 5.582090 2.293733
    5 4.032195 1.830892
    6 6.784617 2.742086
```

strengejacke:
Strange, we have some code to back-transform data for quite a long time:
https://github.com/easystats/insight/blob/654252dca5697b147f2f6df9796a26178742fcd3/R/utils_get_data.R#L51

And this should recognize column names of the model frame:
https://github.com/easystats/insight/blob/654252dca5697b147f2f6df9796a26178742fcd3/R/utils_get_data.R#L794

[...]

PR #691, opened 2022-11-26
revise_get_data

strengejacke:
See #689

Make sure this PR doesn't break other packages:
- [x] datawizard
- [x] bayestestR
- [x] parameters
- [x] performance
- [x] effectsize
- [x] modelbased
- [x] correlation
- [x] report
- [x] see
- [x] easystats

# Breaks

### New behaviour

``` r
library(parameters)
m <- lm(mpg ~ factor(cyl) + hp, mtcars)
parameters::parameters_type(m)
#>      Parameter      Type        Link        Term Variable Level
#> 1  (Intercept) intercept        Mean (Intercept)     <NA>  <NA>
#> 2 factor(cyl)6   unknown        <NA>        <NA>     <NA>  <NA>
#> 3 factor(cyl)8   unknown        <NA>        <NA>     <NA>  <NA>
#> 4           hp   numeric Association          hp       hp  <NA>
#>   Secondary_Parameter Secondary_Type Secondary_Link Secondary_Term
#> 1                <NA>             NA             NA             NA
#> 2                <NA>             NA             NA             NA
#> 3                <NA>             NA             NA             NA
#> 4                <NA>             NA             NA             NA
#>   Secondary_Variable Secondary_Level Tertiary_Parameter
#> 1                 NA              NA                 NA
#> 2                 NA              NA                 NA
#> 3                 NA              NA                 NA
#> 4                 NA              NA                 NA
```

### Expected behaviour

``` r
library(parameters)
m <- lm(mpg ~ factor(cyl) + hp, mtcars)
parameters::parameters_type(m)
#>      Parameter      Type        Link        Term Variable Level
#> 1  (Intercept) intercept        Mean (Intercept)     <NA>  <NA>
#> 2 factor(cyl)6    factor  Difference        cyl6      cyl     6
#> 3 factor(cyl)8    factor  Difference        cyl8      cyl     8
#> 4           hp   numeric Association          hp       hp  <NA>
#>   Secondary_Parameter Secondary_Type Secondary_Link Secondary_Term
#> 1                <NA>             NA             NA             NA
#> 2                <NA>             NA             NA             NA
#> 3                <NA>             NA             NA             NA
#> 4                <NA>             NA             NA             NA
#>   Secondary_Variable Secondary_Level Tertiary_Parameter
#> 1                 NA              NA                 NA
#> 2                 NA              NA                 NA
#> 3                 NA              NA                 NA
#> 4                 NA              NA                 NA
```


R/get_data.R
```diff
@@ -64,6 +64,11 @@ get_data <- function(x, ...) {
         vars <- c(vars, find_offset(x))
         dat <- dat[, intersect(unique(vars), colnames(dat)), drop = FALSE]
       }
+      # remove response for random effects
+      if (effects == "random") {
+        resp <- find_response(x, combine = FALSE)
+        dat <- dat[setdiff(colnames(dat), resp)]
```
vincentarelbundock, line 70: probably missing a comma here

vincentarelbundock: I pushed a commit with the missing comma

R/get_data.R
```diff
@@ -67,7 +67,7 @@ get_data <- function(x, ...) {
       # remove response for random effects
       if (effects == "random") {
         resp <- find_response(x, combine = FALSE)
-        dat <- dat[setdiff(colnames(dat), resp)]
+        dat <- dat[, setdiff(colnames(dat), resp)]
```
strengejacke, line 70: I think this doesn't work, as you need the drop argument, while my approach always returned columns:

``` r
iris[setdiff(colnames(iris), "Species")] |> head()
#>   Sepal.Length Sepal.Width Petal.Length Petal.Width
#> 1          5.1         3.5          1.4         0.2
#> 2          4.9         3.0          1.4         0.2
#> 3          4.7         3.2          1.3         0.2
#> 4          4.6         3.1          1.5         0.2
#> 5          5.0         3.6          1.4         0.2
#> 6          5.4         3.9          1.7         0.4
x <- iris[c(1, 5)]
x[setdiff(colnames(x), "Species")] |> head()
#>   Sepal.Length
#> 1          5.1
#> 2          4.9
#> 3          4.7
#> 4          4.6
#> 5          5.0
#> 6          5.4
x[, setdiff(colnames(x), "Species")] |> head()
#> [1] 5.1 4.9 4.7 4.6 5.0 5.4
```

<sup>Created on 2022-11-27 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

strengejacke:
@vincentarelbundock I believe the code should work now, though some/many tests are still failing. Mainly due to environment issues, and I reduced verbosity [here](https://github.com/easystats/insight/pull/691/commits/8c157926d8d11c3bd53bf4c6c3d78d4ae098e685), else, you would get many warning that model data could not be retrieved from environment, but tests pass due to falling back to the old routines.

In general, I think that other functions that changed (like `find_response()`) are now faster and probably even more stable now. `get_data()` is also much faster and the main code is *much* shorter, but the environment issue remains, in particular in tests...

Not sure what we should do with this PR...

vincentarelbundock:
FWIW, I just installed locally and ran the `marginaleffects` tests. I only had to adjust a few things, mostly with hackish re-assignment structures like:

```r
dat <- mtcars
dat$cyl <- factor(dat$cyl)
dat <<- dat
```

There are a couple things I couldn't figure out for `brglm2` and `lme4` models, and I'll try to diagnose those as soon as possible, but that might not be before next week.

Thanks for all your work on this. This feels like a good improvement!

strengejacke:
Ok, we're getting close to have fixed everything, and indeed, mostly necessary changes in tests were intended and not new unexpected behaviours.

One thing which might be a bit tricky is `test-gam.R:154`: the gam is a multivariate response model, and the `model.frame()` contains the variables `y0` and `y1`, but not the original data frame. So `get_data()` fails here. We either may add an exception for multivariate gams and additionally extract the response variables from the model frame, or we just might rely on the former behaviour of `get_data()` for (multivariate) gams.

vincentarelbundock:
I don't have a strong view on that particular case, but my expectation as a user would be for `get_data()` to generate the sort of dataset I could directly plug into the `data` argument of the `mgcv::gam()` (or other fitting) function. Not sure we need `y0` column to fit a new model... But then again, I rarely use those models, so I don't really know what's best.

strengejacke:
@easystats/core-team This PR introduces a breaking change across several packages. The idea behind this PR is discussed in #689 

I see that it is generally a good idea to have a function that reliably returns the original data used to fit the model (what this PR implements with the revision of `get_data()`), but sometimes we need the additional variables from the model frame. **My question is:** Do we need an argument that can trigger the old/former behaviour? Something like `get_data(model, from_environment = TRUE/FALSE) `?

mattansb:
Oh boy, yes, this will 100% break all the standardization in parameters and datawizrd...

Can we have a `keep_raw  =TRUE/FALSE` to return the original or transformed data? (`from_environment` seems to opaque to me...)

vincentarelbundock:
Another option to consider is: Why bother with subsetting columns at all? Now that we just retrieve from the environment, it is essentially costless to return all the columns in the original data, and those columns could be very useful. Example: I call `insight::get_predicted()`, merge back the original data with `insight::get_data()`, and then look at predictions by subgroups of some other variable.

Maybe this complicates the calls to `complete.cases`, but perhaps if the user really wants the full original dataset, then they simply get the incomplete cases too. I would personally use this as default everywhere.

strengejacke:
@vincentarelbundock see `Parameter` column from the example in the initial post. The columns are not on the original data and hence cannot be retrieved from environment. They're only in the model frame. And there are quite some functions across packages that require these columns (from model frame).

strengejacke:
The model frame contains data per _parameter_, while the data contains data per _variable_.

vincentarelbundock:
Right, sorry I wasn't clear. I meant this as a *separate* feature request. 

Say I want to estimate `lm(mpg ~ hp, mtcars)` and want to retrieve all the columns in the original `mtcars` dataset. Right now, there appears to be an undocumented `additional_variables` switch, but it would be nice if we could get all the columns without having to specify them one by one. That raises issues with `complete.cases()`, but perhaps if users want all columns they also want all rows.

mattansb:
> The model frame contains data per parameter, while the data contains data per variable.

Wouldn't the model matrix have data per parameter?

What would the new `get_data` return here?

``` r
m <- lm(mpg ~ hp + exp(hp), mtcars)

insight::get_data(m) |> head()
#>                    mpg  hp hp.1
#> Mazda RX4         21.0 110  110
#> Mazda RX4 Wag     21.0 110  110
#> Datsun 710        22.8  93   93
#> Hornet 4 Drive    21.4 110  110
#> Hornet Sportabout 18.7 175  175
#> Valiant           18.1 105  105

insight::get_modelmatrix(m) |> head()
#>                   (Intercept)  hp      exp(hp)
#> Mazda RX4                   1 110 5.920972e+47
#> Mazda RX4 Wag               1 110 5.920972e+47
#> Datsun 710                  1  93 2.451246e+40
#> Hornet 4 Drive              1 110 5.920972e+47
#> Hornet Sportabout           1 175 1.003539e+76
#> Valiant                     1 105 3.989520e+45
```

<sup>Created on 2022-11-30 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

rempsyc:
> My question is: Do we need an argument that can trigger the old/former behaviour? Something like get_data(model, from_environment = TRUE/FALSE) ?

I think being able to optionally get data from the environment in cases where the data itself is not available from the model would be very useful indeed in those edge cases. Agree with Vincent we might as well get all the data in a first step, and then if needed subset columns in a second step.

strengejacke:
Ok, I think the only issues that now remains (hopefully) is this failing test, which looks strange to me. Cannot see which test fails? Looks like some gam(m) stuff

```
── Error (???): get_predicted ──────────────────────────────────────────────────
Error in `model.frame.default(formula = y ~ 1 + x0 + x1, data = dat2, drop.unused.levels = TRUE)`: variable lengths differ (found for 'x0')
Backtrace:
    ▆
 1. ├─stats::model.frame(formula = y ~ 1 + x0 + x1, data = dat2, drop.unused.levels = TRUE)
 2. └─stats::model.frame.default(...)
 3. ```
 4.

**Comment by vincentarelbundock** [comment]
> Ok, I think the only issues that now remains (hopefully) is this failing test, which looks strange to me. Cannot see which test fails? Looks like some gam(m) stuff
> 
> ```
> ══ Failed tests ════════════════════════════════════════════════════════════════
> ── Error (???): get_predicted ──────────────────────────────────────────────────
> Error in `model.frame.default(formula = y ~ 1 + x0 + x1, data = dat2, drop.unused.levels = TRUE)`: variable lengths differ (found for 'x0')
> Backtrace:
>     ▆
>  1. ├─stats::model.frame(formula = y ~ 1 + x0 + x1, data = dat2, drop.unused.levels = TRUE)
>  2. └─stats::model.frame.default(...)

@strengejacke took me a while to figure out, but it looks like `mgcv::gam` looks in the environment to grab the variables in the formula. It gets the full `x0` vector, but the `dat3` you fed to the `data` argument has only 30 rows, hence the error message of mismatched length.

One solution would be to create a `sim_data()` function to generate the `dat3` dataset at the top of the test file. That would prevent the environment from being polluted by all the `x0`, `x1`, ... objects.

When I tried, it looked like things were working, but I got another `mgcv::gam` about the shape of the outcome `y`. I couldn't figure that one out, but that's because I don't know the `mgcv` package at all, and I expect you would be able to resolve this pretty easily.

**Comment by strengejacke** [comment]
@IndrajeetPatil your latest commit will also remove checks from PRs... I think this is not intended, but I'm not sure how to deal with this?

**Comment by strengejacke** [comment]
ok, after "update branch from main", tests were gone, but after my comment, they started again... Not sure if this is any technical issue, or just temporary.

**Comment by vincentarelbundock** [comment]
Is that warning too aggressive? Is it important to notify users of which method was used to retrieve the dataset?

Removing it would clean up the tests ;)

**Comment by vincentarelbundock** [comment]
`correlation` is fine, I checked it off the list: https://github.com/easystats/correlation/pull/276

**Comment by vincentarelbundock** [comment]
Errors and warnings in the `see` package appear unrelated to this PR. I checked it off the list: https://github.com/easystats/see/actions/runs/3725126830/jobs/6317734043

**Comment by vincentarelbundock** [comment]
`easystats` is an easy one, so I checked it off the list after local test.

**Comment by vincentarelbundock** [comment]
`modelbased` and `report` are a bit different: with this PR installed and the latest version of `parameters`, I don't get any new error. However, I get a bunch of the "can't find the data in the environement" warnings. I'm not sure if this is expected at this stage of development, so @strengejacke will have to let us know if those can be checked off the list.

**Comment by strengejacke** [comment]
> However, I get a bunch of the "can't find the data in the environement" warnings.

I guess this is because in tests data cannot be found unless assigned with `<<-`.

**Comment by strengejacke** [comment]
But it's good that tests don't break, that was intended that `get_data()` falls back to the old behaviour.

**Comment by vincentarelbundock** [comment]
The `get_data()` function no longer returns a “factor” attribute:

``` r
library(insight)
mod <- lm(mpg ~ factor(am), mtcars)
dat <- get_data(mod)
attr(dat, "factor") |> print()
    NULL
```

strengejacke:
> The `get_data()` function no longer returns a “factor” attribute:
> 
> ```r
> library(insight)
> mod <- lm(mpg ~ factor(am), mtcars)
> dat <- get_data(mod)
> attr(dat, "factor") |> print()
>     NULL
> ```

No, it didn't seem to be necessary, except for `get_datagrid()`, so I moved the code to `get_datagrid.visualization_matrix()`.

Should we add back those attributes by default?

vincentarelbundock:
I did use them, but I can probably work around this. I'll check how much work it is and will come back if there a real problem.

Thanks for the clarification!

R/get_datagrid.R
```diff
@@ -936,6 +924,31 @@ get_datagrid.datagrid <- get_datagrid.visualisation_matrix
...
+    cleaned_terms <- vapply(logical_expressions, all.vars, character(1))
+    for (i in cleaned_terms) {
+      if (is.numeric(data[[i]])) {
+        attr(data[[i]], "logical") <- TRUE
+      }
+    }
+    attr(data, "logicals") <- cleaned_terms
+  }
+
   data
```
strengejacke, line 952: @vincentarelbundock here's the code, we could add it to the new `get_data()` function, like at the end of `.get_data_from_environment()`.

vincentarelbundock: Thanks for the pointer! Feels like you probably had the right intuition here. The vast majority of people who call `get_data()` won't need this, so it probably makes sense to pay the overhead only when it is useful: when building a grid.

vincentarelbundock: Just wanted to confirm that I'm fine without this. Thanks again!

vincentarelbundock:
Anything you would like me to do to help move this along? I'd be happy to spend a couple hours on this if you need it.

(This is self-interested, as this causes a couple of critical bugs downstream where we silently get incorrect numerical results. Sorry to put pressure!!)

strengejacke:
I think there's not much to do here for now.

- [ ] We need to check if this PR causes any problems with the two remaining packages (and if so, add `source = "frame"` in `get_data()` calls)
- [ ] Decide whether we add factor-attributes by default

Else, this is ready to merge, I guess. Once CRAN submission reopens, we need to submit all packages, and _insight_ as last package in that queue, to avoid too many breaking stuff.

vincentarelbundock:
Cool. As I said in the comments, I think your intuition was right all along and that we should NOT assign variable class attributes in `get_data()`. This should just be a function to retrieve the original data.

I'll take a look at the two remaining packages as soon as possible.

IndrajeetPatil:
Since this PR is touching many files, I have pushed changes to tests in the PR itself instead of the `main` to avoid merge conflicts.

vincentarelbundock:
@strengejacke I opened a pull request with a few very minor changes to the test suite of `modelbased`. I don't have merge rights there, so someone else will need to merge, but this should resolve all outstanding problems with `modelbased`: https://github.com/easystats/modelbased/pull/220

vincentarelbundock:
errors are unrelated; no warning that I can see. Merging.

Thanks for all your work on this @strengejacke . This is a massive improvement!

PR #692, opened 2022-12-06
bug: get_predicted_se with incomplete MM

vincentarelbundock:
This PR comments out dangerous code in the calculation of SEs for predicted values. I think it is very important to merge this as soon as possible because the current version of `insight` produces incorrect numerical results.

``` r
library(insight)
set.seed(12345)
n <- 500
x <- sample(1:3, n, replace = TRUE)
y <- rnorm(n)
z <- ifelse(x + y + rlogis(n) > 1.5, 1, 0)
dat <- data.frame(x = factor(x), y = y, z = z)
m <- glm(z ~ x + y, family = binomial, data = dat)
nd <- head(dat, 2)

# good
predict(m, type = "link", newdata = nd, se.fit = TRUE)$se.fit

            1         2 
    0.1924812 0.2117471 

# bad
get_predicted(m, data = head(dat, 2), ci = 0.95, predict = "link") |> data.frame()

       Predicted        SE     CI_low   CI_high
    1 -0.1884105 0.2948762 -0.7663572 0.3895363
    2  1.6547188 0.3374601  0.9933091 2.3161285
```

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Type: Package
 Package: insight
 Title: Easy Access to Model Information for Various Model Objects
-Version: 0.18.8.2
+Version: 0.18.8.3
 Authors@R: 
     c(person(given = "Daniel",
              family = "Lüdecke",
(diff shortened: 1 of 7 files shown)


vincentarelbundock:
Note that this PR involves a couple small regressions in terms of functionality, but they are important because the current version of `insight` produces what I am pretty sure are incorrect results.

Issue #693, opened 2022-12-06
`get_modelmatrix()` padding is broken, with consequences for `get_predicted_se()`


vincentarelbundock:
Padding code in `get_modelmatrix()` does not work when `data` does not include all factor levels. This ended up being a big problem because it causes incorrect results in `get_predicted_se()` (different than the ones already reported).

The problem is that we really need *all* the factor levels when padding with the goal of feeding it to `stats::model.matrix`. Not just a `min_levels`

I will open a PR but demonstrate here. Notice that we are missing a bunch of columns `x2`, `w3`, `w4` in the outcome. Then in `get_predicted_se()`, we subset the variance-covariance matrix to match this amputated model matrix and try to calculate standard errors. But those SEs are totally incorrect.

``` r
library(insight)
set.seed(12345)
n <- 500
x <- sample(1:3, n, replace = TRUE)
w <- sample(1:4, n, replace = TRUE)
y <- rnorm(n)
z <- ifelse(x + y + rlogis(n) > 1.5, 1, 0)
dat <- data.frame(x = factor(x), w = factor(w), y = y, z = z)
m <- glm(z ~ x + w + y, family = binomial, data = dat)
nd <- head(dat, 2)
get_modelmatrix(m, data = nd)
```

      (Intercept) x3 w2          y
    3           1  0  0 -0.6100753
    4           1  1  0  1.1160246

strengejacke:
Could you also update the news file? :-)

vincentarelbundock:
Done :)

PR #694, opened 2022-12-06
Issue #693

vincentarelbundock:
#693


vincentarelbundock:
Merging because this is an important bug that affects correctness of numerical results, and because the test failures only affect a few machines, and seem unrelated to this PR.

PR #702, opened 2022-12-21
Add missing inv-trans

mattansb:
Were missing for `log2` and `log10`


mattansb:
Errors are unrelated to this change.

PR #705, opened 2023-01-05
Support mmrm

strengejacke:
See https://github.com/easystats/parameters/issues/821


strengejacke:
@bbolker In our automatic tests run by Github Actions, we often get an error when testing on older R versions, which is related to _glmmTMB_:

```
Error in .Call("FreeADFunObject", ptr, PACKAGE = DLL) : 
  "FreeADFunObject" not available for .Call() for package "glmmTMB"
Calls: <Anonymous>
```

Any ideas how to avoid this? Run tests that include _glmmTMB_ only on R 4 or higher?

PR #706, opened 2023-01-06
`get_data(mod, additional_variables = TRUE)`

vincentarelbundock:
Retrieve all columns from the original data:

``` r
library(insight)
mod <- lm(mpg ~ hp, data = mtcars)
get_data(mod, additional_variables = TRUE) |> head(3)
```

                   mpg cyl disp  hp drat    wt  qsec vs am gear carb
    Mazda RX4     21.0   6  160 110 3.90 2.620 16.46  0  1    4    4
    Mazda RX4 Wag 21.0   6  160 110 3.90 2.875 17.02  0  1    4    4
    Datsun 710    22.8   4  108  93 3.85 2.320 18.61  1  1    4    1

### DIFF
--- a/R/get_data.R
+++ b/R/get_data.R
@@ -106,7 +106,7 @@ get_data <- function(x, ...) {
       dat <- .recover_data_from_environment(x)
       # additional variables? Some models, like plm::plm(), have an "index"
       # slot in the model call with further variables
-      if (!is.null(additional_variables)) {
+      if (!is.null(additional_variables) && !isTRUE(additional_variables)) {
         vars <- c(vars, additional_variables)
       }
       # add response, only required if "find_variables()" does not already
@@ -129,7 +129,9 @@ get_data <- function(x, ...) {
         if (!is.null(model_call$subset)) {
           vars <- c(vars, all.vars(model_call$subset))
         }
-        dat <- dat[, intersect(unique(vars), colnames(dat)), drop = FALSE]
+        if (!isTRUE(additional_variables)) {
+          dat <- dat[, intersect(unique(vars), colnames(dat)), drop = FALSE]
+        }
       }
       # remove response for random effects
       if (effects == "random") {
@@ -246,7 +248,7 @@ get_data.default <- function(x, source = "environment", verbose = TRUE, ...) {
   }
 
   # try to recover data from environment
-  model_data <- .get_data_from_environment(x, source = source, verbose = verbose)
+  model_data <- .get_data_from_environment(x, source = source, verbose = verbose, ...)
 
   # fall back to extract data from model frame
   if (is.null(model_data)) {
@@ -329,7 +331,7 @@ get_data.mhurdle <- function(x, verbose = TRUE, ...) {
 #' @export
 get_data.mjoint <- function(x, source = "environment", verbose = TRUE, ...) {
   # try to recover data from environment
-  model_data <- .get_data_from_environment(x, source = source, verbose = verbose)
+  model_data <- .get_data_from_environment(x, source = source, verbose = verbose, ...)
 
   if (!is.null(model_data)) {
     return(model_data)
@@ -367,7 +369,8 @@ get_data.geeglm <- function(x,
     x,
     effects = effects,
     source = source,
-    verbose = verbose
+    verbose = verbose,
+    ...
   )
 
   if (!is.null(model_data)) {
@@ -403,7 +406,8 @@ get_data.gee <- function(x,
   model_data <- .get_data_from_environment(x,
     effects = effects,
     source = source,
-    verbose = verbose
+    verbose = verbose,
+    ...
   )
 
   if (!is.null(model_data)) {
@@ -440,7 +444,8 @@ get_data.rqss <- function(x,
   model_data <- .get_data_from_environment(x,
     component = component,
     source = source,
-    verbose = verbose
+    verbose = verbose,
+    ...
   )
 
   if (!is.null(model_data)) {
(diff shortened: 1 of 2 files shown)


vincentarelbundock:
Personally, I think that this option should be the default. Subsetting the columns made sense when we didn't want to retrieve and backtransform each of them individually, but now it's very cheap to return the full original data (there's no memory benefit since we manipulate the big data set internally anyway).

In this PR, I just pushed through the ellipsis to allow `additional_variables=TRUE`, but I'd be curious to know what @strengejacke and others think about the defaults.

DominiqueMakowski:
I'm not sure, I'd suppose most of the time when one extracts the data from a model the interest lies in the data used by the model, and any additional content seems like a loss in efficiency and memory usage, which can become problematic in the case of huge datasets

vincentarelbundock:
> I'm not sure, I'd suppose most of the time when one extracts the data from a model the interest lies in the data used by the model, and any additional content seems like a loss in efficiency and memory usage, which can become problematic in the case of huge datasets

Fair enough. But we are already manipulating the full dataset inside `get_data` before subsetting the columns. AFAICT this means that the full dataset is already in memory, so my guess is that there is actually 0 cost in terms of efficiency.

R/get_data.R
```diff
@@ -106,7 +106,7 @@ get_data <- function(x, ...) {
       dat <- .recover_data_from_environment(x)
       # additional variables? Some models, like plm::plm(), have an "index"
       # slot in the model call with further variables
-      if (!is.null(additional_variables)) {
+      if (!is.null(additional_variables) && !isTRUE(additional_variables)) {
```
strengejacke, line 109: What is if `additional_variables` is `TRUE`? Then `TRUE` is added to `vars`?

strengejacke:
I think by default we should _not_ retrieve the complete data frame _by default_. Say we have a large survey with >100k rows and >400+ columns, and only 8 variables are used to fit the model. The returned data frame is much too large. Furthermore, if I recall correctly, R has copy-on-modify, i.e. we indeed multiply large data frames in memory, where mostly not required.

vincentarelbundock:
To be clear, I don’t think this is a big deal and I’m happy to live with the majority choice here, but I think it is both more cpu *and* memory efficient to *not* subset in this case. This is probably because your argument about copy-on-modify also applies when we subset the columns:

``` r
library(dqrng)
library(insight)
library(microbenchmark)

dat <- data.frame(matrix(dqrnorm(2e8), ncol = 10000))
dim(dat)
```

    [1] 20000 10000

``` r
mod <- lm(X1 ~ X2, data = dat)

bench::mark(
    get_data(mod),
    get_data(mod, additional_variables = TRUE),
    check = FALSE
)
```

    # A tibble: 2 × 6
      expression                                      min   median itr/se…¹ mem_al…²
      <bch:expr>                                 <bch:tm> <bch:tm>    <dbl> <bch:by>
    1 get_data(mod)                                1.66ms   1.96ms     461.   1.14MB
    2 get_data(mod, additional_variables = TRUE)   1.57ms   1.93ms     456. 514.41KB
    # … with 1 more variable: `gc/sec` <dbl>, and abbreviated variable names
    #   ¹​`itr/sec`, ²​mem_alloc

strengejacke:
> because your argument about copy-on-modify also applies

Yes, because in most real-world-scenarios, we won't get the exact data, e.g. due to case-wise-deletion etc. If we then keep all variables, we have two data sets in memory (which, honestly, might not be a big deal, unless we call `get_data()` multiple times).

- I think the returned data should contain as many observations as used by the model (so we remove in-complete cases)
- when we keep _all_ variables, extracting valid cases is **much more** difficult, because you need to apply `complete.cases()` or `na.omit()` only to the data that was used to fit the model.
- I don't want "bloated" data sets.

That's why I think the _default_ should be to return only variables used in the model.

vincentarelbundock:
Sounds good, thanks!

strengejacke:
btw, if `additional_variables` is no character vector, but `TRUE`, we could possibly simplify:

instead
```r
      # recover data frame from environment
      dat <- .recover_data_from_environment(x)
      # additional variables? Some models, like plm::plm(), have an "index"
      # slot in the model call with further variables
      if (!is.null(additional_variables) && !isTRUE(additional_variables)) {
        vars <- c(vars, additional_variables)
      }
```

we use

```r
      # recover data frame from environment
      dat <- .recover_data_from_environment(x)
      if (isTRUE(additional_variables)) {
        return(dat)
      }
      # additional variables? Some models, like plm::plm(), have an "index"
      # slot in the model call with further variables
      if (!is.null(additional_variables) && !isTRUE(additional_variables)) {
        vars <- c(vars, additional_variables)
      }
```

WDYT?

vincentarelbundock:
Currently, we use the `vars` object below to do smart `complete.cases()` and return just the right rows, even if there are more columns (with `NA`s). I'm not sure if what you suggest would return all the rows...

strengejacke:
No, it wouldn't. I just thought you wanted an option to return the full original data... Else, we could leave the code as is (or add a second argument `full_data` or similar)

vincentarelbundock:
Ah, gotcha. Thanks, but I think the current options work fine for me for now. Maybe something to keep in mind in the future if some requests it. Would be pretty easy to add, but no need to make the code more complex unless there's concrete demand.

strengejacke:
I'm now planning submissions for _parameters_ and _performance_, which must be on CRAN before _insight_, so we can submit _insight_, soon. The other packages were not broken by the changes in `get_data()`, right? I remember you have tested them in separate PRs.

vincentarelbundock:
Wonderful!

Yes, I did test everything, so crossing fingers we shouldn't have much (if any) issues (famous last words).

vincentarelbundock:
@strengejacke 

Maybe wait a bit on `insight`. I'm finding some corner cases in `get_data()` and will try to iron those out in the near future. Will ping you when I'm done.

strengejacke:
Ok!

PR #710, opened 2023-01-09
`rstanarm` tests cleanup

vincentarelbundock:


vincentarelbundock:
@strengejacke this completes the things I had on my TODO list for `insight`. Tests seem to pass on the few "important" CI machines, but I have to admit that I never really understand what is going on with all those workflows...

vincentarelbundock:
@IndrajeetPatil Maybe we could try installing `TMB` from source, as I do elsewhere: 

https://github.com/vincentarelbundock/marginaleffects/blob/main/.github/workflows/check.yaml#L49

IndrajeetPatil:
Good idea. I am installing both `{TMB}` and `{Matrix}` from source now. Let's see if that makes any difference.

IndrajeetPatil:
R CMD check is passing now!  🎉

https://github.com/easystats/insight/actions/runs/3878381591/jobs/6614483687

strengejacke:
Yeah, finally! 🥳

PR #715, opened 2023-01-25
Fix `has_single_value` when only `NA` and `na.rm = TRUE`

etiennebacher:
New behavior:
``` r
library(insight)
has_single_value(c(NA, NA), na.rm = TRUE)
#> [1] FALSE
```


strengejacke (approved):

strengejacke:
Thanks!

PR #724, opened 2023-02-21
options for `format_value()` : drop leading 0, minus and plus signs

### LINKED ISSUE #711: options for `format_value()` : drop leading 0, minus and plus signs
We use `format_value()` a bunch in *see* and `visualisation_recipe()` for formatting labels on graphs. It would be useful to include some more number formatting options to make these prettier. Three that come to mind:

1. Drop leading zeros (e.g., from `visualisation_recipe.see_easycorr_matrix`)
2. Format negative signs as minus signs vs hyphens
3. Format positive signs as plus or figure space vs nothing

For 2 or 3, cf. https://github.com/r-lib/scales/pull/366

strengejacke:
Fixes #711

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Type: Package
 Package: insight
 Title: Easy Access to Model Information for Various Model Objects
-Version: 0.19.0.1
+Version: 0.19.0.2
 Authors@R: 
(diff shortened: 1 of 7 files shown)


bwiernik:
Awesome thanks!

PR #728, opened 2023-03-10
find_offset.fixest

vincentarelbundock:


strengejacke:
See lintr check:

> file=R/find_offset.R,line=49,col=65,[string_boundary_linter] Use !is.na(x) & startsWith(x, string) to detect a fixed initial substring, or, if missingness is not a concern, just startsWith. Doing so is more readable and more efficient.

😁

vincentarelbundock:
This is a bad suggestion, because it fails on `NULL`.

(I dislike automatic linting.)

vincentarelbundock:
Haha

Sure, but this is only executed once, so the most you can ever get in terms of wall time is 6 milliseconds...

strengejacke:
Sorry, I'm too conditioned by the lintr-integration in VS Code ;-)

PR #736, opened 2023-03-17
find_offset for fixest minor fix

vincentarelbundock:


R/find_offset.R
```diff
@@ -55,5 +55,10 @@ find_offset <- function(x) {
 
   offset <- clean_names(offset)
 
-  offset
+  # sometimes we get an empty list (e.g., fixest with iris dataset)
+  if (isTRUE(length(offset) == 0)) {
```
strengejacke, line 59: Or

```
if (length(offset)) return offset else return NULL
```

(sorry for bad formatting, typing from smartphone)

PR #742, opened 2023-03-19
Small memory improvement for `trim_ws()` 

etiennebacher:
I realised in https://github.com/easystats/datawizard/pull/384 that `trim_ws()` uses a lot of memory compared to its base equivalent. Using `perl = TRUE` reduces it:

``` r
new_trim_ws <- function(x) {
  gsub("^\\s+|\\s+$", "", x, perl = TRUE)
}

test <- rep("  Some text. ", 10000)

bench::mark(
  base = trimws(test),
  old = insight::trim_ws(test),
  new = new_trim_ws(test),
  iterations = 1000
)
#> # A tibble: 3 × 6
#>   expression      min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr> <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 base         4.99ms   5.44ms      168.  156.34KB    0.337
#> 2 old          6.06ms   6.68ms      134.    1.33MB    0.135
#> 3 new          3.97ms   4.72ms      179.   78.17KB    0.359
```

<sup>Created on 2023-03-19 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

### DIFF
--- a/R/utilities.R
+++ b/R/utilities.R
@@ -46,7 +46,7 @@ trim_ws <- function(x, ...) {
 
 #' @export
 trim_ws.default <- function(x, ...) {
-  gsub("^\\s+|\\s+$", "", x)
+  gsub("^\\s+|\\s+$", "", x, perl = TRUE)
 }
 
 #' @rdname trim_ws


strengejacke:
Good to know! I thought using `perl = TRUE` will always be less sufficient.

etiennebacher:
Actually what are the differences between `base::trimws` and `insight::trim_ws`? I see that the regex used are different but is there a reason for this?

strengejacke:
Not sure whether we weren't aware and/or for minor performance reasons...

Here's a comparison for different lengths of the input vector:

plot

Memory allocation seems to be identical for old and new `trim_ws()`. Only base R differs

plot2

etiennebacher:
That's strange, I now find that memory allocation is identical but using `perl = TRUE` is slightly faster. 

Well, feel free to merge or close this one

strengejacke:
It's an improvement, even if small, so merging. Thanks!

PR #744, opened 2023-03-21
Clean names for "weak instruments" in ivreg

etiennebacher:
Related to https://github.com/easystats/performance/pull/560

I don't know where to fix the format so that it prints `< .001` instead of the real p-value (see output in `performance`'s PR)


strengejacke:
It's in `.format_p_values()`, I have fixed it.

PR #746, opened 2023-03-23
fix mclogit support

strengejacke:


R/get_statistic.R
```diff
@@ -972,6 +972,47 @@ get_statistic.mlogit <- function(x, ...) {
   }
 }
 
+#' @export
+get_statistic.mclogit <- function(x, ...) {
+  if (requireNamespace("mclogit", quietly = TRUE)) {
```
IndrajeetPatil, line 977: Why not use `insight::check_if_installed()` here to avoid control flow?

strengejacke: was too lazy, just copy/pasted ;-)

PR #747, opened 2023-03-23
Bump minimum needed R version to `R 3.6`

IndrajeetPatil:
cf. https://github.com/easystats/easystats/issues/304

- `R 4.3` release is a month away (21st April).
- We don't test `R 3.5` in our CI/CD either.


strengejacke:
I think we can ask @vincentarelbundock here and his R version dependencies. But for _modelsummary_, _parameters_ and _performance_ already require R 3.6, so maybe only _marginaleffects_ is affected?

strengejacke:
https://cran.r-project.org/web/packages/marginaleffects/index.html

No, not affected. I think we can safely bump requirements to R3.6, and maybe @vincentarelbundock you do the same for _modelsummary_?

vincentarelbundock:
> No, not affected. I think we can safely bump requirements to R3.6, and maybe @vincentarelbundock you do the same for _modelsummary_?

Sounds good. Thanks for the heads up!

IndrajeetPatil:
@strengejacke Are we backporting anything from `R 3.5` that can be removed? Nothing comes to my mind, at least.

strengejacke:
I don't think so, I think we had all backports in a file that also had this name

strengejacke:
https://github.com/easystats/insight/blob/main/R/backports.R

PR #749, opened 2023-03-27
Support phylolm

strengejacke:


IndrajeetPatil:
Thanks, @strengejacke! 👏

strengejacke:
Not sure if we should add tests. I tested everything locally, but the SUGGEST field would become very big when we actually test every model we support? How should we deal with that?

PR #751, opened 2023-04-07
Test cleanup

etiennebacher:
https://github.com/easystats/easystats/issues/294
https://github.com/easystats/easystats/issues/350
https://github.com/easystats/easystats/issues/358
https://github.com/easystats/easystats/issues/363

@easystats/core-team This is not hard but very tedious so I'd like some help on this (~50% done already, cf commit message to see where I stopped)

In summary:

- [x] remove the `if()` condition by either putting the `skip_` functions at the top of the file or inside `test_that()` depending on the context
- [x] use snapshots instead of manually formatted output -> not sure always done but should be good enough for now
- [x] use `skip_on_os()` instead of defining e.g `.windows` and then calling `.windows` in `if ()`
- [x] add `<pkg>::` where needed because we don't call `library()` in tests anymore (also need to specify the package in `data()` calls). It is sometimes needed to call `library()` but this **shouldn't** be followed by `unloadNamespace()` because there can be errors if we try to unload a package that is used by other packages (even not called with `library()`). Ideally, if all functions are prefixed with `<pkg>::`, having a few calls to `library()` shouldn't be a problem.
- [x] conditions like `.runStanTests` (or sth like this) can be replaced by `skip_on_cran()`. 
- [x] use `skip_if_offline()` everywhere `download_model()` is used


IndrajeetPatil:
Thanks for getting the ball rolling, @etiennebacher. 

I am currently traveling without my laptop and can only help Tuesday onwards.

IndrajeetPatil:
By the time the PR is ready for a merge:

1,000,000 additions
1,000,000 deletions 

😆

etiennebacher:
> By the time the PR is ready for a merge:
>
> 1,000,000 additions
> 1,000,000 deletions

Good luck to the reviewers :stuck_out_tongue_winking_eye:

etiennebacher:
@IndrajeetPatil @strengejacke there are 2-3 errors with win-devel about `brms` but I'm not sure how to handle them. Apart from that, I think this is ready for review

etiennebacher:
@IndrajeetPatil I fixed some issues in random test order that are due to `formula.tools::as.character.formula()` being in the namespace. One alternative would be to use `as.character.default()` instead of `as.character()` in `insight` to bypass this method dispatch issue.

tests/testthat/test-find_formula-data.R
```diff
@@ -1,4 +1,9 @@
-data(mtcars)
+# see https://github.com/georgheinze/logistf/pull/54
```
IndrajeetPatil, line 1: Thanks for making the PR upstream!

IndrajeetPatil (approved):
Thanks again, @etiennebacher, for all the hard work. Having done this for other repos, I know how frustrating and tedious this drudge work is, but, hopefully, it will be a “done once, done for eternity” kind of change.

Daniel is going to buy you a beer when you meet 😄 

P.S. The failing test for mipo is unrelated to this PR. It's also failing on the main.

etiennebacher:
It was actually a bit faster than I thought with 20min here and there, thanks also for the fixes

> Daniel is going to buy you a beer when you meet 😄

Looking forward to it 😄

IndrajeetPatil:
Yeah, some things are straightforward find and replace but fixing tests failing in random order because S3 method dispatches are being overridden is no fun.

PR #758, opened 2023-04-14
Tests failing in random order (again)

### LINKED ISSUE #757: Tests failing in random order (again)
@etiennebacher Do you want to have a look at it?

strengejacke:
Fixes #757


etiennebacher:
I can't reproduce the `test-format_table.R` error on Windows and there's no error message so no idea where it comes from

IndrajeetPatil:
Check succeeds now. I changed the R version from devel to release. I think that was the issue. 
We shouldn't be running this check on R-devel.

strengejacke:
Cool thanks 👍

PR #761, opened 2023-05-02
Fix #760, so `get_data()` can fetch from the correct environment, and add tests

### LINKED ISSUE #760: `get_data()` omits variables created in a non-global environment
I'm not sure if the title is completely correct, but it's what I've deduced so far. Here's a test script:

```r
library(insight)

foo <- function() {
  mtcars$cylinders <- factor(mtcars$cyl)

  fit <- lm(mpg ~ cylinders * disp + hp, data = mtcars)

  return(get_data(fit))
}

cat("Fit inside other environment:\n")
print(names(foo()))

cat("In global environment:\n")

mtcars$cylinders <- factor(mtcars$cyl)

fit <- lm(mpg ~ cylinders * disp + hp, data = mtcars)

print(names(get_data(fit)))
```

This yields the following output:

```
Fit inside other environment:
[1] "mpg"  "disp" "hp"  
In global environment:
[1] "mpg"       "cylinders" "disp"      "hp
```

Notice that `cylinders` is present in both models, but only appears in the `get_data()` output when it's created in the global environment. And yet `get_data()` is called from the same environment as the model is fit in both cases, so it's not like we're doing wacky environment hijinks.

I think the cause may be related to #666, because this problem appeared when I upgraded from 0.18.8 to 0.19.1, and that seems like a relevant change during that time period. I happened to run into the problem in a testthat test case, but as the example above shows, it need not be in a test case to occur.

I haven't tried to narrow this down further, but I can test out specific package versions or changes if that'd help.

capnrefsmmat:
The commit messages should explain the issue, and I've added a comment explaining the `enclos` issue, since it's definitely not obvious.

I've also added tests for three cases:

* data is present in the local environment
* data is in a different environment
* data is in a different environment, and another object of the same name is in the parent environment

All of these fail without the fix. But it's worth adding more if you can think of additional cases, since this is definitely easy to mess up. Are there other cases you'd expect to cause problems?

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -57,7 +57,12 @@ Authors@R:
              family = "Thériault",
              role = "ctb",
              email = "remi.theriault@mail.mcgill.ca",
-             comment = c(ORCID = "0000-0003-4315-6788", Twitter = "@rempsyc")))
+             comment = c(ORCID = "0000-0003-4315-6788", Twitter = "@rempsyc")),
+      person(given = "Alex",
+             fammily = "Reinhart",
+             role = "ctb",
+             email = "areinhar@stat.cmu.edu",
+             comment = c(ORCID = "0000-0002-6658-514X")))
 Maintainer: Daniel Lüdecke <d.luedecke@uke.de>
 Description: A tool to provide an easy, intuitive and consistent
     access to information contained in various R models, like model
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -59,7 +59,7 @@ Authors@R:
              email = "remi.theriault@mail.mcgill.ca",
              comment = c(ORCID = "0000-0003-4315-6788", Twitter = "@rempsyc")),
       person(given = "Alex",
-             fammily = "Reinhart",
+             family = "Reinhart",
              role = "ctb",
              email = "areinhar@stat.cmu.edu",
              comment = c(ORCID = "0000-0002-6658-514X")))
--- a/NEWS.md
+++ b/NEWS.md
@@ -29,6 +29,13 @@
 
 * Fixed issue in `model_info()` for models of class `gamlss`.
 
+* Fixed problems preventing `get_data()` from locating data defined in
+  non-global environments.
+
+* Fixed issue in `get_predicted()` for variables of class numeric matrix created
+  by `scale()`, which were correctly handled only when `get_data()` failed to
+  find the data in the appropriate environment.
+
 # insight 0.19.1
 
 ## New supported models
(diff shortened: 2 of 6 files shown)


capnrefsmmat:
I happened to notice the mgcv test while reviewing the other tests, and it seems to work now. Not sure if that was fixed by my changes or by something else, but I reenabled it and added a couple explanatory comments.

strengejacke:
Thanks a lot, the changes looks good! You may add yourself as contributor to the DESCRIPTION file, if you like. And it would be great if you could also add a news item. :-)

strengejacke:
I think this test failure is new:

```
── Error ('test-get_predicted.R:538:3'): bugfix: used to fail with matrix variables ──
Error: Error: variable 'wt' was fitted with type "numeric" but type "nmatrix.1" was supplied
Backtrace:
    ▆
 1. ├─insight::get_predicted(m) at test-get_predicted.R:538:2
 2. └─insight:::get_predicted.lm(m) at insight/R/get_predicted.R:208:2
 3.   └─insight (local) predict_function(x, data = args$data) at insight/R/get_predicted.R:340:4
 4.     ├─stats::predict(...) at insight/R/get_predicted.R:314:4
 5.     └─stats::predict.lm(...)
 6.       └─stats::.checkMFClasses(cl, m)

[ FAIL 1 | WARN 0 | SKIP 132 | PASS 784 ]
Error: Error: Test failures
Execution halted
```

capnrefsmmat:
Well, this is weird. I installed 0.19.1 on a different machine, without the changes from this branch, and running the code from the test fails:

```
> library(insight)
>   mtcars2 <- mtcars
  mtcars2$wt <- scale(mtcars2$wt)
  m <- lm(mpg ~ wt + cyl + gear + disp, data = mtcars2)
  pred <- get_predicted(m)
Some of the variables were in matrix-format - probably you used
  `scale()` on your data?
  If so, and you get an error, please try `datawizard::standardize()` to
  standardize your data.
Error: variable 'wt' was fitted with type "numeric" but type "nmatrix.1" was supplied
```

But it's clearly passed on CI before.

My guess is that `get_predicted.lm()` should be fixing the classes *after* using `.get_predicted_args()` to get the data, so the correction is applied to the right data frame. But I don't understand how the code worked before, so clearly there is some other problem with the code or the tests that allowed the test to pass when the code fails when run separately.

capnrefsmmat:
Okay, pushed NEWS and DESCRIPTION changes, and I think this is ready to be tested again.

capnrefsmmat:
Dang -- I fixed the DESCRIPTION mistake now.

strengejacke:
Thanks, lgtm! The failing tests are not related to this PR.

strengejacke (approved):

PR #768, opened 2023-05-13
Dangerous `get_data()` behavior with `fixest` models and name conflicts

### LINKED ISSUE #767: Dangerous `get_data()` behavior with `fixest` models and name conflicts
I was not sure if this issue would be better solved from within the `fixest` package or here, but since it is similar to issue #760, I thought I would try here first.

When a `fixest` model is fit with a function that returns the model and `get_data()` is called from the global environment, if an object with the same name and overlapping columns exists, `get_data()` will return that object instead.

This issue does not occur with the normal `glm()` function.

Reproducible example:

```r
library(insight)
library(fixest)

# fit within function
fixest_wrapper <- function(data) {
    data$cylinders <- factor(data$cyl)

    fit <- feglm(mpg ~ cylinders * disp + hp, data = data)

    return(fit)
}

glm_wrapper <- function(data) {
    data$cylinders <- factor(data$cyl)

    fit <- glm(mpg ~ cylinders * disp + hp, data = data)

    return(fit)
}

# use function
global_fixest <- fixest_wrapper(data = mtcars)
global_glm <- glm_wrapper(data = mtcars)

# create name conflict
data <- mtcars[, c('mpg', 'disp')]

cat("global_fixest:\n")
print(names(get_data(global_fixest)))

cat("global_glm:\n")
print(names(get_data(global_glm)))
```

Returns:
```
global_fixest:
[1] "mpg"  "disp"
global_glm:
[1] "mpg"       "cylinders" "disp"      "hp"  
```

Package versions:
```
fixest_0.11.1     insight_0.19.1.10
```

strengejacke:
Fixes #767

### DIFF
--- a/R/get_data.R
+++ b/R/get_data.R
@@ -246,7 +246,7 @@ get_data <- function(x, ...) {
     dat_fixest <- .safe(eval(model_call$data, envir = parent.env(x$call_env)))
     # sanity check - does data from parent env. differ from current extracted
     # data? If so, use data from parent env.
-    if ((dim(dat_fixest)[1] == dim(dat)[1]) && (dim(dat_fixest)[2] != dim(dat)[2])) {
+    if (!is.null(dat_fixest) && (dim(dat_fixest)[1] == dim(dat)[1]) && (dim(dat_fixest)[2] != dim(dat)[2])) {
       dat <- dat_fixest
     }
   }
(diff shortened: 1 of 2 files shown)


strengejacke:
Some tests for different settings needed.

PR #795, opened 2023-08-06
`check_if_installed()` now automatically determines the correct minimum version

### LINKED ISSUE #793: Suggestion of new function: `get_dep_version()`
# Context

`insight::check_if_installed()` is a great alternative to `rlang::is_installed()`.

During the review of my rempsyc paper at JOSS, it was pointed out that for suggested packages, the wrong package versions were checked. I assumed these functions were smart enough to check the DESCRIPTION document for the correct version, but this is not the case. So we need to specify the minimal version manually using the `minimum_version` or `version` arguments.

# Problem and Solution

This is suboptimal because it creates additional maintenance: we have to update the version numbers not only in the DESCRIPTION file but also in every function that depends on those suggested packages (crazy I know!).

For this purpose, I created a function, `get_dep_version()`, that automatically detects the minimal version of a given package from the DESCRIPTION file.

# Reprex:

``` r
rempsyc::get_dep_version(dep = "rstanarm", pkg = "insight")
#> [1] "2.21.1"

rempsyc::get_dep_version(dep = "see", pkg = "modelbased")
#> [1] "0.7.4"
```

<sup>Created on 2023-08-05 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

This is useful because it can be directly integrated with `insight::check_if_installed()`:

``` r
insight::check_if_installed("see", minimum_version = rempsyc::get_dep_version(
  dep = "see", pkg = "modelbased"))
#> Error: Package `see` is installed, but package version `0.7.4` is required.
#>   Please update it by running `install.packages(`see`)`.
```

<sup>Created on 2023-08-05 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

In fact, this could be the default, should we want it. This would allow us to reduce maintenance in all easystats packages by changing the required version number only once: where it belongs in the DESCRIPTION file. Furthermore, making this the default in `insight::check_if_installed()` would show a CLEAR benefit of `insight::check_if_installed()` over `rlang::is_installed()`.

---

### LINKED ISSUE #794: `check_if_installed()` errors when specifying more than one package and version
Working on #793 and realized an existing problem with a fresh GitHub install when specifying more than one `package` and `minimum_version`:

``` r
library(insight)
packageVersion("insight")
#> [1] '0.19.3.1'

# package & minimum_version works
check_if_installed("datawizard", minimum_version = "0.8.0")

# package > 1 works
check_if_installed(c("datawizard", "rstanarm"))

# package > 1 & minimum_version > 1 doesn't work
check_if_installed(c("datawizard", "rstanarm"), minimum_version = c("0.8.0", "2.21.1"))
#> Error in if (pkg == "base") file.path(.Library, "base") else if (isNamespaceLoaded(pkg)) getNamespaceInfo(pkg, : the condition has length > 1
```

Whereas `rlang::is_installed()` works as expected:

``` r
rlang::is_installed(c("datawizard", "rstanarm"), version = c("0.8.0", "2.21.1"))
#> [1] TRUE
```

<sup>Created on 2023-08-06 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

This was not discovered earlier because there are no examples or tests testing this scenario.

Edit: will fix it in the PR

rempsyc:
`check_if_installed()` now automatically checks the package DESCRIPTION file to determine the correct minimum version required.

---

Closes #793 and closes #794

### DIFF
--- a/R/check_if_installed.R
+++ b/R/check_if_installed.R
@@ -160,8 +160,7 @@ get_dep_version <- function(dep, pkg = utils::packageName()) {
   out <- lapply(dep, function(x) {
     dep.string <- grep(paste0("\n", x), suggests.list, value = TRUE, fixed = TRUE)
     dep.string <- unlist(strsplit(dep.string, ">", fixed = TRUE))
-    out <- gsub("[^0-9.]+", "", dep.string[2])
-    out
+    gsub("[^0-9.]+", "", dep.string[2])
   })
   out <- unlist(out)
   if (all(is.na(out))) {
--- a/R/check_if_installed.R
+++ b/R/check_if_installed.R
@@ -46,7 +46,7 @@ check_if_installed <- function(package,
                                quietly = FALSE,
                                prompt = interactive(),
                                ...) {
-  is_installed <- sapply(package, requireNamespace, quietly = TRUE)
+  is_installed <- vapply(package, requireNamespace, quietly = TRUE, FUN.VALUE = TRUE)
   what_is_wrong <- what_you_can_do <- NULL
 
   if (is.null(minimum_version)) {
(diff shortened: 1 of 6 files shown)


rempsyc:
I guess it's difficult to add tests for this function given that the required packages may or may not be available due to the R CMD check hard workflow where those packages will not be available.

R/check_if_installed.R
```diff
@@ -128,3 +153,18 @@ print.check_if_installed <- function(x, ...) {
     print_color(paste0("- ", names(x)[!x], collapse = "\n"), "red")
   }
 }
+
+get_dep_version <- function(dep, pkg = utils::packageName()) {
```
bwiernik, line 157: I suggest we call this `.get_dep_version()` given that it's an internal function.

bwiernik (approved):
Looks good to me. Made a small comment about naming, otherwise I'm fine to merge once checks pass

rempsyc:
Remaining failing checks appear unrelated to this PR

bwiernik:
Thanks!

PR #796, opened 2023-08-13
Fix edge cases in `.get_dep_version()`

rempsyc:
Fix edge cases where there are more than one `grep` matches (e.g. packages `boot` and `bootES`) and when there is a required version for the first package of the Suggests list which created a bug.


rempsyc:
Errors are unrelated to this PR

PR #809, opened 2023-09-22
code clean

strengejacke:
Related to:

- https://github.com/easystats/easystats/issues/318
- https://github.com/easystats/easystats/issues/334
- https://github.com/easystats/easystats/issues/187

### DIFF
--- a/R/find_formula.R
+++ b/R/find_formula.R
@@ -564,14 +564,14 @@ find_formula.RM <- find_formula.MANOVA
 find_formula.gls <- function(x, verbose = TRUE, ...) {
   ## TODO this is an intermediate fix to return the correlation variables from gls-objects
   fcorr <- x$call$correlation
-  if (!is.null(fcorr)) {
+  if (is.null(fcorr)) {
+    f_corr <- NULL
+  } else {
     if (inherits(fcorr, "name")) {
       f_corr <- attributes(eval(fcorr))$formula
     } else {
       f_corr <- parse(text = safe_deparse(fcorr))[[1]]
     }
-  } else {
-    f_corr <- NULL
   }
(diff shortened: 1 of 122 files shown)


IndrajeetPatil:
Thanks a lot, Daniel, for working on the clean-up! ❤️

strengejacke:
I'd say when these checks pass, we can merge for now. And address remaining things in a new PR.

R/find_formula.R
```diff
@@ -2,69 +2,67 @@
...
 #'
-#'    - `conditional`, the "fixed effects" part from the model (in the
-#'      context of fixed-effects or instrumental variable regression, also
-#'      called *regressors*) . One exception are `DirichletRegModel` models
-#'      from \pkg{DirichletReg}, which has two or three components,
-#'      depending on `model`.
+#'  - `conditional`, the "fixed effects" part from the model (in the
+#'    context of fixed-effects or instrumental variable regression, also
+#'    called *regressors*) . One exception are `DirichletRegModel` models
+#'    from **DirichletReg**, which has two or three components,
```
IndrajeetPatil, line 22: What motivates this change? `\pkg{}` -> `**`?

I would prefer it we instead use `` `{pkg}` `` because then pkgdown can link to that package's website (if available). You don't need to revert this change. Just for future reference.

IndrajeetPatil:
Strict workflows are failing.

R/find_formula.R
strengejacke: `\pkg{}` was just rendered as bold text, that's why I replaced it with the related markdown tag.

strengejacke: We could use regex to replace, shouldn't be too hard to check occurrences

PR #813, opened 2023-09-26
`find_transformation` incorrectly detects power transformations

### LINKED ISSUE #812: `find_transformation` incorrectly detects power transformations
The `find_transformation()` function is looking for a regex  `"I\\((.*)\\^\\s*2\\)"` to identify a power transformation

* transformations of the *response variable* in a formula do not need to be protected by `I()`; the special interpretation of operators `*+^/` only applies to the right-hand side of a formula
* even so, I don't understand the regex here: we need an `I()`-protected quantity raised to a power that ends in 2??
* I don't know how flexible a regex you want, exactly; should probably allow for `^` followed by a quantity that might be in parentheses, `+-`, possible `/`, digits ... decimal points (in case some decides to raise to the 0.5), ... ??? It might be useful to include the rules in the documentation ...

```
library(insight)
find_transformation(lm(Sepal.Length^(1/2) ~ Species, data = iris))
## [1] "identity"
find_transformation(lm(Sepal.Length^2~ Species, data = iris))
## [1] "identity"
find_transformation(lm(I(Sepal.Length^(1/2)) ~ Species, data = iris))
## NULL
 find_transformation(lm(I(Sepal.Length^3) ~ Species, data = iris))
## NULL
find_transformation(lm(I(Sepal.Length^2) ~ Species, data = iris))
## [1] "power"
```

strengejacke:
Fixes #812

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Type: Package
 Package: insight
 Title: Easy Access to Model Information for Various Model Objects
-Version: 0.19.5.3
+Version: 0.19.5.4
 Authors@R: 
     c(person(given = "Daniel",
              family = "Lüdecke",
--- a/R/find_terms.R
+++ b/R/find_terms.R
@@ -188,11 +188,8 @@ find_terms.mipo <- function(x, flatten = FALSE, ...) {
   # protect "-1"
   f$conditional <- gsub("(-1|- 1)(?![^(]*\\))", "#1", f$conditional, perl = TRUE)
 
-  ## TODO: this pattern currently fails for this response example:
-  # lm(Sepal.Length^-0.5 ~ Species, data = iris)
-  # "Sepal.Length^-0.5" is considered as subtraction, thus it is split into
-  # "Sepal.Length^" and "0.5".
-
+  # This regular expression matches any of the characters *, +, :, |, -, or /,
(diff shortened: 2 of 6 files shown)


bwiernik:
The regex should also accommodate:

1. Decimal exponents 
2. Using `**` instead of `^`

strengejacke:
Should work.

@bbolker Most use-cases should work now, except negative power, which is an issue in `find_terms()`. Having a `-` in the response is considered as subtraction, like success/trials, e.g. `cbind(a-b)`. Thus, the response is split into two part, assuming that left and right side of `-` are variables.

``` r
m <- lm(Sepal.Length^-0.5 ~ Species, data = iris)
insight::find_terms(m)
#> $response
#> [1] "Sepal.Length^" "0.5"          
#> 
#> $conditional
#> [1] "Species"
```

Not sure how to fix this? The relevant code starts here:
https://github.com/easystats/insight/blob/b87c1c3f1b4f9b9ee0380890c069a52a570f3dee/R/find_terms.R#L191

@bwiernik @bbolker is it even possible to return the power as function? See:
https://github.com/easystats/insight/blob/b87c1c3f1b4f9b9ee0380890c069a52a570f3dee/R/get_transformation.R#L61

bwiernik:
```
list(transformation = function(x, power) x^power, inverse = function(x, power) x^(1/power))
```

PR #829, opened 2023-11-02
Use withr to not change global state

strengejacke:
Fixes https://github.com/easystats/easystats/issues/388

### DIFF
--- a/tests/testthat/test-backticks.R
+++ b/tests/testthat/test-backticks.R
@@ -1,122 +1,110 @@
-iris$`a m` <- iris$Species
-iris$`Sepal Width` <- iris$Sepal.Width
-dat <<- iris
-m <- lm(`Sepal Width` ~ Petal.Length + `a m` * log(Sepal.Length), data = dat)
-m2 <- lm(`Sepal Width` ~ Petal.Length + `a m`, data = dat)
+skip_if_not_installed("withr")
 
-test_that("text_remove_backticks", {
-  d <- data.frame(Parameter = names(coef(m2)), Estimate = unname(coef(m2)), stringsAsFactors = FALSE)
-  expect_equal(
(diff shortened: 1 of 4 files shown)


strengejacke:
@IndrajeetPatil Can I add multiple `test_that()` code chunks into one `withr::*()` call?
Like:

```r
withr::with_environment(
  new.env(),
  m <- lm(...)
  test_that("1", {
    ...
  })
  test_that("2", {
    ...
  })
  test_that("3", {
    ...
  })
)

**Comment by IndrajeetPatil** [comment]
Yes, but IINM you'd need to do something like this:


```r
withr::with_environment(
  new.env(),
  code = {
    m <- lm(...)
    test_that("1", {
      ...
    })
    test_that("2", {
      ...
    })
    test_that("3", {
      ...
    })
  }
)
```

strengejacke:
Cool! This should resolve our scoping issues, and probably also the random test order thing.

IndrajeetPatil:
You could do that, or maybe a better approach is to create a new environment in each test and test the expectations within this scope.

For example, see https://github.com/r-lib/withr/blob/0c5254ebc74590e80cc0056303d74b049b943920/tests/testthat/test-namespace.R#L70-L87

strengejacke:
> You could do that, or maybe a better approach is to create a new environment in each test and test the expectations within this scope.

This is what I would normally do, but if you look at the test structure of _insight_, e.g. https://github.com/easystats/insight/blob/main/tests/testthat/test-lm.R, I would need to re-fit the model every time again for each test then (in those cases where we use `<<-`, which we want to avoid).

IndrajeetPatil:
Maybe add back the removed "random test order" check?

strengejacke:
I think that check should work once we addressed the scoping issues using _withr_, so I'll do that after the withr-work is done.

PR #841, opened 2023-12-23
Added support for `mgcv::gaulss` for `find_formula` and `find_variables`

### LINKED ISSUE #840: Add family == "gaulss" to `find_formula` and `find_variables` for `mgcv::gam`
Hi,

`marginaleffects::datagrid` by @vincentarelbundock uses `find_variables` which calls to `find_formula`. Currently, `find_formula.gam` don't handle `.gam_family(x) == "gaulss"`, resulting in `find_variables` failing to grab the correct variables.

```r
# Problem
library(insight)
library(marginaleffects)
library(mgcv)

dat <- gamSim(1, n = 100, dist = "normal", scale = 2)
b <- gam(list(y ~ s(x0) + s(x1) + s(x2), ~ s(x3)), family = gaulss(), data = dat)

find_formula(b) # Nameless list
find_variables(b) # response = "", no conditional

datagrid(model = m) # 0*0 dataframe
```

I propose adding support for `mgcv::gaulss`. At least for just `marginaleffects::datagrid`, it involves adding a case for `gaulss`  to `R/find_formula.R/find_formula` and updating `R/find_variables.R/find_variables.R`

```r
# R/find_formula.R/find_formula
find_formula.gam <- function(x, verbose = TRUE, ...) {
  f <- tryCatch(stats::formula(x), error = function(x) NULL)

  if (!is.null(f)) {
    if (is.list(f)) {
      mi <- .gam_family(x)
      if (!is.null(mi) && mi$family == "ziplss") {
        # handle formula for zero-inflated models
        f <- list(conditional = f[[1]], zero_inflated = f[[2]])
      } else if (mi$family == "Multivariate normal") {
        # handle formula for multivariate models
        r <- lapply(f, function(.i) deparse(.i[[2]]))
        f <- lapply(f, function(.i) list(conditional = .i))
        names(f) <- r
        attr(f, "is_mv") <- "1"
      } else if (mi$family == "gaulss") {  # <------------ Add this
        f <- list(conditional = f[[1]], scale = f[[2]])
      }
    } else {
      f <- list(conditional = f)
    }
  }

  .find_formula_return(f, verbose = verbose)
}

# R/find_variables.R/find_variables.R
find_variables <- function(x,
                           effects = "all",
                           component = "all",
                           flatten = FALSE,
                           verbose = TRUE) {
  effects <- match.arg(effects, choices = c("all", "fixed", "random"))
  component <- match.arg(component, choices = c("all", "conditional", "zi", "zero_inflated", "dispersion", "instruments",
 "smooth_terms",  "scale")) # <- Added term at the end
...
```

Now `marginaleffects::datagrid` works properly.
```r
datagrid(model = m)
```

I have 3 questions:
1) Would it make sense to name the "scale" parameter as "sigma" due to this being a normal distribution? Or would it make sense to name it "scale" since it is a member of the location–scale exponential family? Naming it "scale" maybe better since we can add other extended gam family like `mgcv::gumbls` as well and we wouldn't have to update find_variables with extra notations.
2) Should we use this chance to add other gam extended families as well?
3) If you need my help changing other functions to accommodate this change for {insight} let me know and I'll investigate.

Great package, and thank you so much.

hhp94:
Fixes #840 

Added support for `mgcv::gaulss` in `find_formula` and `find_variables`.

For `R/find_formula.R/find_formula.gam`, i'm pretty sure the `is.null` is supposed to be outside of the if `mi$familly == x` checks. Other than that, just a straight forward case of checking if the family is "gaulss" and if yes then extract not only the conditional variables for the location parameter but also the variables for the scale parameter.

For `R/find_variables.R/find_variables`, have to update the choices to also have "scale".

Added tests for this change. Sorry if there are any mistakes, this is my first PR ever


strengejacke:
Thanks a lot! I turned the `if...else` into a `switch` statement (suggested by _lintr_), and also addressed some other lintr-issues that were outside the scope of this PR (but still flagged, because they appeared in the files that have been changed - _lintr_ checks modified files in PRs, so lintr-issues that existed before but were not spotted are now mentioned).

strengejacke:
Thanks, I think remaining failing tests are not related to this PR

hhp94:
The rstats community is great. Merry Christmas! I'll lookingto `model_info` and the link function related functions soon.

PR #843, opened 2024-01-24
Detect multivariate vgams/vglms

B0ydT:
I thought I might as well update this here so that the ggeffects function can use `model_info` for everything including checking for multivariate models. Works for both `is_multivariate` and `model_info`.

Updated one test which which expected `is_multivariate == FALSE`, I imagine as a placeholder - it has two outcome variables.

I figure this is a cleaner solution but let me know if you'd rather leave as is.

strengejacke/ggeffects#432

### DIFF
--- a/R/is_multivariate.R
+++ b/R/is_multivariate.R
@@ -59,5 +59,5 @@ is_multivariate <- function(x) {
     return(isTRUE(x@extra$multiple.responses))
   }
 
-  return(FALSE)
+  FALSE
 }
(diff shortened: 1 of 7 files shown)


strengejacke:
Thanks! Will look closer at your PRs later, but at first glance they look good!

B0ydT:
Thanks! I've taken a look at some for the the failed checks, and they don't look related to my changes, but please let me know if I've missed something or can help in any way.

strengejacke:
Yes, should be good! Thanks!

PR #861, opened 2024-03-20
updates for glmmTMB changes

bbolker:
There are two changes here:

1. account for the fact that Gaussian dispersion is parameterized as log-SD rather than log-variance now
2. change starting value for `theta` to avoid a singular fit (caused by slight changes in numerical stability due to Gaussian reparameterization)

also some minor spelling corrections (sorry)

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Type: Package
 Package: insight
 Title: Easy Access to Model Information for Various Model Objects
-Version: 0.19.9.1
+Version: 0.19.9.2
 Authors@R: 
     c(person(given = "Daniel",
              family = "Lüdecke",
--- a/NEWS.md
+++ b/NEWS.md
@@ -5,6 +5,8 @@
 * Function like `find_variables()` or `clean_names()` now support multi-membership
   formulas for models from *brms*.
 
+* Updated tests to work with the latest changes in *glmmTMBM 1.1.9*.
+
 # insight 0.19.9
 
 ## New supported models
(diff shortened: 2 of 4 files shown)


strengejacke:
This _may_ also break code in _performance_ (but I'm planning a new release the next days anyway). I think functions like `icc()` or `r2_nakagawa()` should be affected, if variance components change. On the other hand, if the breaking change is:

> account for the fact that Gaussian dispersion is parameterized as log-SD rather than log-variance now

then tests shouldn't probably be changed, but rather we want to fix `insight::get_variance()`, so that indeed the variance, not SD, is returned again? I think that makes more sense, in order to get correct results for r2/icc again.

bbolker:
I think I fixed that one too?

Yes, these tests should be conditional.

I don't see any need to require version 1.1.9 in DESCRIPTION.

bbolker:
Have to run, will respond later, but hopefully `get_variance` would be using accessor methods such as `sigma()` or `VarCorr` that would be robust to these internal changes?

strengejacke:
> I think I fixed that one too?

Yeah, the PR still used glmmTMB 1.1.8, that's why tests are failing now.

strengejacke:
> Have to run, will respond later, but hopefully `get_variance` would be using accessor methods such as `sigma()` or `VarCorr` that would be robust to these internal changes?

Ok, I'll also take a look.

strengejacke:
r2/icc are identical between glmmTMB version 1.1.8 and 1.1.9, so no need for action here.

strengejacke:
Thanks, looks good!

PR #885, opened 2024-06-09
Implement `download_model()` using `{httr2}` and add tests

IndrajeetPatil:


R/download_model.R
```diff
@@ -10,6 +10,8 @@
 #'   Optional, and should only be used in case the repository-URL is
 #'   changing. By default, models are downloaded from
 #'   `https://raw.github.com/easystats/circus/master/data/`.
+#' @param extension File extension. Default is `.rda`.
```
IndrajeetPatil, line 13: In case we ever start saving data in some other format.

R/download_model.R
```diff
@@ -19,43 +21,49 @@
...
+#' try(download_model("non_existent_model"))
+#' }
+#'
 #' @export
-download_model <- function(name, url = NULL) {
-  .download_data_github(name, url)
+download_model <- function(name,
+                           url = "https://raw.github.com/easystats/circus/master/data/",
+                           extension = ".rda",
+                           verbose = TRUE) {
```
IndrajeetPatil, line 34: Verbosity is useful here to control output in tests.

tests/testthat/test-download_model.R
```diff
@@ -0,0 +1,20 @@
+httptest2::with_mock_dir("download-model-success", {
```
IndrajeetPatil, line 1: None of these tests need `skip_if_offline()` because they are mocking HTTP requests.

tests/testthat/download-model-failure/raw.github.com/easystats/circus/master/data/xyz.rda.R
strengejacke, line 1: Why do we need these files? Are they used in tests?

tests/testthat/setup.R
strengejacke, line 1: What happens if the package is not available?

tests/testthat/test-download_model.R
strengejacke, line 1: Shouldn't we add a `skip_if_not_installed()` here, as we do in other tests for other packages, too?

tests/testthat/download-model-failure/raw.github.com/easystats/circus/master/data/xyz.rda.R
IndrajeetPatil: Yes, they are mocked responses.

strengejacke (approved):

strengejacke:
Do we need both httptest2 and httr2 to use `download_model()`, or is httr2 sufficient?

IndrajeetPatil:
httptest2 is for tests, httr2 provides the client. 

The user needs only the latter.

strengejacke:
thanks! we need to change this (httr -> httr2) in our tests where we use `download_model()`, that's why I'm asking which packages should go into `skip_if_not_installed()`.

IndrajeetPatil:
You are right! 

Sorry, missed it in this PR. I should have just removed _httr_ from DESCRIPTION.

IndrajeetPatil:
SGTM.

Issue #889, opened 2024-06-14
Validate `get_variance()` against remaining families


strengejacke:
As a follow-up of #877

Following families need validation or don't yet work:

- [ ] betabinomial
- [ ] hurdle models
- [ ] zero-inflated models
- [ ] glmmTMB::compois
- [ ] glmmTMB::genpois

strengejacke:
We have following deviation to MuMIn:

- **Computation of fixed effects variance**: https://github.com/easystats/insight/issues/877#issuecomment-2163052765
- **Computation of distribution-specific variance for negative binomial models (negbin/nbinom1)**: https://github.com/easystats/insight/issues/877#issuecomment-2165551707
- **Computation of Sigma for _glmmTMB_**: https://github.com/easystats/insight/issues/877#issuecomment-2165696977

However, the things we do differently are in line with the code in the Supplement 2 of Nakagawa. Thus, I would say we go with the current implementation of `get_variance()`. I added *lots* of tests to validate against external evidence of correct results.

strengejacke:
@bbolker `null_model()` probably doesn't calculate the correct null model for zero-inflation models. What would be the correct null model in the next example?

``` r
data(fish, package = "insight")
m1 <- glmmTMB::glmmTMB(
  count ~ child + camper + (1 | persons),
  ziformula = ~ child + camper + (1 | persons),
  data = fish,
  family = poisson()
)

summary(insight::null_model(m1))
#>  Family: poisson  ( log )
#> Formula:          count ~ (1 | persons)
#> Zero inflation:         ~child + camper + (1 | persons)
#> Data: fish
#> 
#>      AIC      BIC   logLik deviance df.resid 
#>   1808.7   1829.8   -898.3   1796.7      244 
#> 
#> Random effects:
#> 
#> Conditional model:
#>  Groups  Name        Variance Std.Dev.
#>  persons (Intercept) 0.6808   0.8251  
#> Number of obs: 250, groups:  persons, 4
#> 
#> Zero-inflation model:
#>  Groups  Name        Variance Std.Dev.
#>  persons (Intercept) 1.069    1.034   
#> Number of obs: 250, groups:  persons, 4
#> 
#> Conditional model:
#>             Estimate Std. Error z value Pr(>|z|)    
#> (Intercept)   1.5875     0.4173   3.804 0.000142 ***
#> ---
#> Signif. codes:  0 '***' 0.001 '**' 0.01 '*' 0.05 '.' 0.1 ' ' 1
[...]

PR #893, opened 2024-06-14
Test get_variance for zero-inflation models

strengejacke:
Related to #889

### DIFF
--- a/R/compute_variances.R
+++ b/R/compute_variances.R
@@ -831,7 +831,28 @@
   cvsquared <- tryCatch(
     {
       if (faminfo$link_function == "tweedie") {
+        # Tweedie models ------------------------------------------------------
+        # ---------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


strengejacke:
What would be expected as R2 result for zero inflated models in comparison to non zero inflated? Would R2 be higher or lower for simple Poisson as opposed to zero inflated Poisson, for example?

strengejacke:
I think, `mean()` is better. The important parameters are `mu` and `p`. When we take the average `mu` and `p` from the predictions, it is similar to taking the `mean()` at the end.

```r
p <- stats::predict(model, type = "zprob")
mu <- stats::predict(model, type = "conditional")
pvar <- (1 - p) * (mu + p * mu^2)
mean(pvar)

# similar to
p <- mean(stats::predict(model, type = "zprob"))
mu <- mean(stats::predict(model, type = "conditional"))
pvar <- (1 - p) * (mu + p * mu^2)
pvar
```

Issue #929, opened 2024-09-11
Discrepancy in residual variance between VarCorr() and get_variance_residual()


matschmitz:
There seems to be a discrepancy between the residual variance calculated using `brms::VarCorr()` and `insight::get_variance_residual()` for a brms model. Apologies if I'm missing something obvious.

```r
library(brms)    # v2.21.0
library(insight) # v0.20.4.3

mdl <- brm(mpg ~ hp + (1 | cyl), data = mtcars, seed = 123)
# Family: gaussian 
#  Links: mu = identity; sigma = identity 
# Formula: mpg ~ hp + (1 | cyl) 
#  Data: mtcars (Number of observations: 32) 
#  Draws: 4 chains, each with iter = 2000; warmup = 1000; thin = 1;
#         total post-warmup draws = 4000
# 
# Multilevel Hyperparameters:
#   ~cyl (Number of levels: 3) 
#               Estimate Est.Error l-95% CI u-95% CI Rhat Bulk_ESS Tail_ESS
# sd(Intercept)     4.75      2.55     1.33    11.21 1.00     1192     1719
# 
# Regression Coefficients:
#           Estimate Est.Error l-95% CI u-95% CI Rhat Bulk_ESS Tail_ESS
# Intercept    24.79      3.50    17.26    31.25 1.00     1454     1545
# hp           -0.03      0.02    -0.07    -0.00 1.00     2036     2145
# 
# Further Distributional Parameters:
#       Estimate Est.Error l-95% CI u-95% CI Rhat Bulk_ESS Tail_ESS
# sigma     3.29      0.46     2.53     4.32 1.00     2045     2157

VC <- VarCorr(mdl)
# $cyl
# $cyl$sd
#           Estimate Est.Error     Q2.5    Q97.5
# Intercept 4.751544  2.549082 1.334246 11.21103
# 
# 
# $residual__
# $residual__$sd
# Estimate Est.Error     Q2.5    Q97.5
# 3.290936 0.4609515 2.531565 4.321534

VC$residual__$sd[1, 1]^2 # Residual variance
# [1] 10.83026

get_variance_residual(mdl)
# var.residual 
#      7.85012 
```

mariobecerra:
Seems to be related to the new version. I'm running an earlier version (0.19.8) and I do get the same result in both cases (10.83026).

strengejacke:
``` r
library(brms)    # v2.21.0
library(insight) # v0.20.4.3

[...]

Issue #634, opened 2022-09-15
Export `get_dispersion()`


IndrajeetPatil:
In addition to assisting `get_auxiliary()` it should also be useful on its own.

PR #938, opened 2024-10-02
Prepare Release 1.0.0: Check {insight} API / documentation on API

### LINKED ISSUE #937: Check {insight} API / documentation on API
See https://github.com/easystats/easystats/issues/73

To add support for models, especially for _parameters_ and including all features, we need following methods:

- `insight::find_formula()`
- `insight::model_info()`
- `insight::get_parameters()`
- `insight::get_statistic()`
- `insight::find_statistic()`
- `insight::get_varcov()`
- `insight::get_df()`

In other packages, these are optional, when the above methods work fine:
- `parameters::ci()`
- `parameters::p_value()`
- `parameters::standard_error()` **or** `insight::get_varcov()`

The _simple_ way is just adding a method for `parameters::model_parameters()`, however, than this model-class is not supported by _all_ functions we offer in the _easystats_ eco-system.

I think we need a vignette on the API, explaining how to add methods/support for new models, similar to what @mattansb has done here: https://easystats.github.io/effectsize/articles/effectsize_API.html

Maybe we can split those vignettes into two (or more?) parts:

- What to do to add new methods for extracting/tidying parameters? That would only require a method for `model_parameters()`.
- What to do to add new methods for some _easystats_ support? (I'm thinking of `check_model()`, which seems to be often used by users) This would probably require some more methods - must check.
- What to do to add new methods for full _easystats_ support? This would need methods in _insight_, maybe also _parameters_ or _bayestestR_.

strengejacke:
Fixes #937

TODO
- [x] https://github.com/easystats/bayestestR/issues/536
- [x] We need to allow methods for `find_statistic()`, so other can add their classed.
- [ ] https://github.com/easystats/insight/issues/929
- [x] https://github.com/easystats/insight/issues/634

Anything else?


strengejacke:
Tagging @vincentarelbundock - we're preparing some stable 1.0.0 releases in the near future. Anything in the API design or urgent features that should be taken into account for that release?

vincentarelbundock:
Everything is fine for me, thanks for the tag!

Stable release is an **excellent** idea. Great news!

PR #942, opened 2024-10-21
Improvement in how get_transformation() deals with power transformations

### LINKED ISSUE #941: Improvement in how get_transformation() deals with power transformations
### Background
The behaviour of get_transformation() to determine the transformation and inversion functions of transformed responses in a handful of hard-coded steps is surprising to me.
This is related to the issue #940, but might impede other uses as well.

### Suggestion
I suggest to replace this code:
```
  else if (transform_fun == "power") {
    trans_power <- .safe(gsub("\\(|\\)", "", gsub("(.*)(\\^|\\*\\*)\\s*(\\d+|[()])", 
      "\\3", find_terms(x)[["response"]])))
    if (is.null(trans_power)) {
      trans_power <- "2"
    }
    out <- switch(trans_power, `0.5` = list(transformation = function(x) x^0.5, 
      inverse = function(x) x^2), `3` = list(transformation = function(x) x^3, 
      inverse = function(x) x^(1/3)), `4` = list(transformation = function(x) x^4, 
      inverse = function(x) x^0.25), `5` = list(transformation = function(x) x^5, 
      inverse = function(x) x^0.2), list(transformation = function(x) x^2, 
      inverse = sqrt))
  }
```
, that is rows 31 to 43 in insight::get_transformation(), with something like this code:
```
  else if (transform_fun == "power") {
    trans_power <- .safe(gsub("\\(|\\)", "", gsub("(.*)(\\^|\\*\\*)\\s*(\\d+|[()])", 
      "\\3", find_terms(x)[["response"]])))
    if (is.null(trans_power)) {
      trans_power <- "2"
    }
    if (trans_power==0) {
out <- NULL } else{
out<-list(
  transformation = function(x) x ^ as.numeric(trans_power), inverse= function(x) x ^ (as.numeric(trans_power)^-1)
)
}
  }
```
The only changes are in the `out <- `but I included the whole `if`. I hope it's clear.

To the best of my understanding, that should catch and invert any power function. 
I don't know if the functions in `out` created this way will be OK to use downstream, and I can't easily test it from here, but I'm guessing it can be solved by the package experts.

### LINKED ISSUE #943: Transformed responses in performance_aic
### Background
There are some gaps in the ability of performance_aic() to detect and correct transformed responses.
It is of course very hard to cover every possible case, but the documentation does promise a lot:
> performance_aic() correctly detects transformed response and, unlike stats::AIC(), returns the "corrected" AIC value on the original scale. 

### Suggested action

1. I suggest the function should return a warning when we know something is up with the response variable, but the transformation can't be fully identified and corrected. I'm talking about the cases where insight::find_transformation returns NULL.
something like:
`>The transformation and inverse-transformation functions could not be determined. The AIC has been determined without correction.`

2. If Insight::find_transformation==NULL should have this role of triggering warnings in other functions, it would be nice if it covered all suspicious cases. The regex  `any(grepl("I\\((.*)\\)", x))`  is beyond me, but it seems a divided response is neither caught nor corrected. See below.
Wouldn't it be ideal if the occurrence of anything that is not a number or a named object (or a transformation that is handled) would return a NULL  transformation?
```mtcars$mpg_trans<-mtcars$mpg/0.7
a<-aov(formula=mpg/0.7 ~ hp, data=mtcars) # transformation is overt, and should be corrected
b<-aov(formula=mpg_trans ~ hp, data=mtcars) # transformation is hidden in the data, and can't be corrected
c<-aov(formula=I(mpg/0.7) ~ hp, data=mtcars) # same as a, but is identified as NULL transformation

find_transformation(a) # "identity"
find_transformation(b) # "identity"
find_transformation(c) # NULL

performance_aicc(a) # 204.923
performance_aicc(b) # 204.923
performance_aicc(c) # 204.923

### PR DESCRIPTION
**Pull request by strengejacke**
Fixes #941
Fixes #943

### CODE REVIEW

**Comment by strengejacke** [comment]
Do you have examples for the first two points? (and whether they should fail/warn/...)

**Comment by JoakimStefansson** [comment]
**On the first point**, I'm not at all sure this is a problem. I never work with negative dependent variables, and I don't know what would be the desired behaviour for someone who was. **It's probably fine the way it is.**
I just brought it up because I felt some responsibility for claiming to cover any power transformation, when there are limits in applicability, even if the limits were already there.
Here's an example, anyway:
```
dafr<-data.frame("y"=mtcars$mpg*-1, "x"=mtcars$hp) # data that is inappropriate for power transformations (impossible to restore through inversion)
exp<-2
eval(parse(text=paste0("mod<-lm(y^", as.character(exp), "~x, data=dafr)"))) # applying a power transformation
gt<-insight::get_transformation(mod) # no warning at this stage
gt$inverse(gt$transformation(dafr$y)) # old inversion
((dafr$y)^exp)^(exp^(-1)) # new inversion
# when exp==2, the old and new inversion are abs(y)
# when exp==3, the old and new inversion are all NaNs
# in no case is the inversion == y
performance::performance_aic(lm(y^2~x, data=dafr)) # becomes Inf, which should disqualify the model to most users
performance::performance_aic(lm(y^3~x, data=dafr)) # becomes Inf, which should disqualify the model to most users

```

**Comment by JoakimStefansson** [comment]
**On the second point**, here's an example:
```
mod<-lm(mpg^0~hp, data=mtcars)
```
Of course this is a stupid model if it is spelled out, but it may very well happen if the exponent is the result of a function.
The old get_transformation seems to allow inversion to the original data, but this is because sqrt(y^2) is performed according to the default exponent 2. It looks OK, but for the wrong reason.
```
gt<-insight::get_transformation(mod)
gt$inverse(gt$transformation(mtcars$mpg))==mtcars$mpg
```
With the new get_transformation, the inverted transformed y will all be the number 1.
```
(mtcars$mpg^0)^(0^(-1))==1
performance::performance_aic(mod)
```
A valid-looking AIC is returned, which could be dangerous. 
However, summary(mod) warns of an essentially perfect fit, which should deter users.

**What I think should happen**
The other invalid or unintelligible transformations makes get_transformation() return NULL, so I think y^0 should also make it return NULL. 
We can set `out<-NULL`  from inside the power-block and keep the structure of the function. It won't activate the verbose alert, but it's probably fine, since y^0 should be an unusual situation (and a mistake).

**Comment by strengejacke** [comment]
> On the first point, I'm not at all sure this is a problem. I never work with negative dependent variables, and I don't know what would be the desired behaviour for someone who was. It's probably fine the way it is.

Ah, I see. I don't think this is a problem, because negative values are all positive on the transformed scale, and back-transforming is no problem then:

``` r
data(iris)
m <- lm(Sepal.Length^2 ~ Species, data = iris)
fun <- insight::get_transformation(m)

# fake response, with negative values, *as in the data*
y <- -3:3

# values of original response *after* transformation
transformed_y <- fun$transformation(y)
transformed_y
#> [1] 9 4 1 0 1 4 9

# back-transforming transformed y
fun$inverse(transformed_y)
#> [1] 3 2 1 0 1 2 3
```

<sup>Created on 2024-10-22 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

strengejacke:
> performance::performance_aic

I see again, it's problematic for the corrected IC. Maybe again throw a warning that corrected IC could not be calculated then...

PR #973, opened 2024-11-20
Partial fail in new Box-Cox-handling in insight 0.99.0.15

### LINKED ISSUE #970: Partial fail in new Box-Cox-handling in insight 0.99.0.15
First of all, thank you for the great work with these packages, and for implementing changes so quickly. 
I may well be too early, testing functions that aren't official yet.
But anyway, this is how insight is working right now.

```
library(easystats)
packageVersion("insight") # ‘0.99.0.15’
packageVersion("performance") # ‘0.12.4.5’
packageVersion("easystats") # ‘0.7.3.2’

a <- aov(formula = (mpg ^ 1.3 - 1) / 1.3 ~ hp, data = mtcars)
insight::find_transformation(a) # "box-cox" - correct
insight::get_transformation(a) # also has a correct output

a <- aov(formula = (mpg ^ -1.3 - 1) / -1.3 ~ hp, data = mtcars)
insight::find_transformation(a) # "power" - not correct
debugonce(insight::get_transformation)
insight::get_transformation(a)
# Error in !is.null(trans_power) && trans_power != 0 : 
#   'length = 2' in coercion to 'logical(1)'
# In addition: Warning message:
#   In doTryCatch(return(expr), name, parentenv, handler) :
#   NAs introduced by coercion
```
The error in `get:transformation()` appears to be caused by the misdirection from `find_transformation()`.
When running `.extract_power_transformation(model)` instead of `.extract_scale_denominator(model)` on the formula, we get `c(NA, 1.3)` instead of the single value that the next line wants to evaluate.

But there may be more to this bug. While in debugging mode inside `.get_transformation()`, I evaluated `.extract_scale_denominator(model)`, and it becomes **1.3** but I think it ought to be **-1.3**. 
Something happens in the lapply call between line 21 and 31 of `insight:::.get_variables_list()`, when `f$response` goes from  `"(mpg^-1.3 - 1)/-1.3"` to `"(mpg^-1.3 - 1)" "1.3" `.

strengejacke:
Fixes #970

### DIFF
--- a/R/find_terms.R
+++ b/R/find_terms.R
@@ -196,6 +196,7 @@ find_terms.mipo <- function(x, flatten = FALSE, ...) {
 
   # protect "-1"
   f$conditional <- gsub("(-1|- 1)(?![^(]*\\))", "#1", f$conditional, perl = TRUE)
+  f$response <- gsub("(-1|- 1)(?![^(]*\\))", "#1", f$response, perl = TRUE)
 
   # This regular expression matches any of the characters *, +, :, |, -, or /,
   # unless they are preceded by a ^ and followed by a closing parenthesis ).
@@ -239,7 +240,9 @@ find_terms.mipo <- function(x, flatten = FALSE, ...) {
       trim_ws(unlist(strsplit(f$conditional[need_split], " ", fixed = TRUE), use.names = FALSE))
     )
   }
+
   f$conditional <- gsub("#1", "-1", f$conditional, fixed = TRUE)
+  f$response <- gsub("#1", "-1", f$response, fixed = TRUE)
 
   # reorder, so response is first
   compact_list(f[c(length(f), 1:(length(f) - 1))])
--- a/R/find_terms.R
+++ b/R/find_terms.R
@@ -257,7 +257,7 @@ find_terms.mipo <- function(x, flatten = FALSE, ...) {
   error <- utils::capture.output(print(f[[3]][i][[1]]))
   f[[3]][i] <- NULL
   f[[3]] <- f[[3]][[2]]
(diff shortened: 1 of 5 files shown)


strengejacke:
Yes

JoakimStefansson:
It doesn't seem to be. Should I open a new issue?

strengejacke:
Works for me:

``` r
library(insight)
packageVersion("insight") # ‘0.99.0.15’
#> [1] '0.99.0.16'

a <- aov(formula = (mpg ^ 1.3 - 1) / 1.3 ~ hp, data = mtcars)
insight::find_transformation(a) # "box-cox" - correct
#> [1] "box-cox"
insight::get_transformation(a) # also has a correct output
#> $transformation
#> function(x) (x^1.3-1) / 1.3
#> <environment: 0x000001af044791c0>
#> 
#> $inverse
#> function(x) exp(log(1 + 1.3 * x) / 1.3)
#> <environment: 0x000001af044791c0>

a <- aov(formula = (mpg ^ -1.3 - 1) / -1.3 ~ hp, data = mtcars)
insight::find_transformation(a)
#> [1] "box-cox"
insight::get_transformation(a)
#> $transformation
#> function(x) (x^-1.3-1) / -1.3
#> <environment: 0x000001af068b0778>
#> 
#> $inverse
#> function(x) exp(log(1 + -1.3 * x) / -1.3)
#> <environment: 0x000001af068b0778>
```

Try `easystats::install_latest(force = TRUE)` to update to most recent package versions.

strengejacke:
`source="development"`, that's the default, you don't need to specify that. But sometimes, if we forget to increase version numbers in a PR, `easystats::install_latest()` won't install newer versions, because version numbers were the same. `force = TRUE` _always_ installs from r-universe and does not check version numbers.

PR #979, opened 2024-12-12
Docs for get datagrid

### LINKED ISSUE #978: Update `get_datagrid(<marginaleffects>)` for structural changes in 0.24.0
Previously the `get_datagrid(<marginaleffects>)` function relied on the `attr(mfx, "newdata")` attribute, but in the newer version `{marginaleffects}` this attributed seems to no long be available when the `newdata=` argument isn't explicitly set.

@vincentarelbundock Is this intentional? If so, I'm trying to figure out what the best course of action is here to get the relevant column names (those that contain the data id-ing the rows, and not the statistical results).

Is the new `attr(mfx, "original") |> colnames()` appropriate to use here? Or `attr(mfx, "newdata_variable_class") |> names()`?

Please advise.

---

- [ ] Need to also write (better?) tests to catch such changes?

mattansb:
Closes #978


mattansb:
Errors don't seem to be related to this change.

strengejacke:
> Errors don't seem to be related to this change.

@vincentarelbundock this seems to be related to https://github.com/easystats/parameters/issues/1048, right?

Failing test is:

```
── Failure ('test-marginaleffects.R:34:3'): marginaleffects ────────────────────
Names of parameters::model_parameters(x) ('rowid', 'Parameter', 'Comparison', 'Coefficient', 'SE', 'Statistic', 'p', 'S', 'CI', 'CI_low', 'CI_high', 'Predicted', 'Species', 'Petal.Length', 'Sepal.Width') don't match 'rowid', 'Parameter', 'Coefficient', 'SE', 'Statistic', 'p', 'S', 'CI', 'CI_low', 'CI_high', 'Predicted', 'Species', 'Petal.Length', 'Sepal.Width'
```

So we make this conditional on _marginaleffects 1.0.0_?

strengejacke:
@mattansb The check-hard error is related to this PR:

`get_datagrid.emmGrid` needs examples run conditionally.

vincentarelbundock:
Yes, I think that's right. The current `main` version from Github should work, in principle.

strengejacke:
hm, strange, we're importing the current main via the `Remotes` field.

strengejacke:
We don't expect the `Comparison` column:

``` r
packageVersion("marginaleffects")
#> [1] '0.24.0.6'

m <- lm(Sepal.Width ~ Species * Petal.Length, data = iris)

x <- marginaleffects::slopes(m,
  variables = "Petal.Length",
  newdata = insight::get_datagrid(m, by = "Species")
)

setdiff(
  colnames(parameters::model_parameters(x)),
  c(
    "rowid", "Parameter", "Coefficient", "SE", "Statistic", "p",
    "S", "CI", "CI_low", "CI_high", "Predicted", "Species",
    "Petal.Length", "Sepal.Width"
  )  
)
#> [1] "Comparison"
```

<sup>Created on 2024-12-13 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

strengejacke:
Or is the `Comparison` column now expected? Looks like the test in _parameters_ expects it. Strange, this test wasn't touched for months.

vincentarelbundock:
Hmm, I don't know what is going on. Maybe these kinds of tests are too strict? We don't want a break whenever a package moves columns around, unless it is a "material" change.

strengejacke:
Agree, I think its purpose is to check whether the s-value column was renamed into "S". will fix test.

strengejacke:
Ah, no, that was the test in _parameters_ :-D

But, most important: the column `Comparison`, is it included in the output (and expected) or not? I remember you saying that some columns should be removed in `print()`, but not in the underlying data frame.

vincentarelbundock:
In the current CRAN, it is *not* include. In the future CRAN versions and on Github, it *will* be included.

strengejacke:
ok, I fixed the text. Shouldn't be crucial, as it's not tested on CRAN anyway.

mattansb:
Thanks for taking care of this @strengejacke 🙏

PR #994, opened 2025-01-15
`include_random` in `get_datagrid()` now works for nested random effects

### LINKED ISSUE #981: get_predicted() with nested random effects
There are two problems when using the `get_predicted()` function on a model containing nested random effect terms.

First, the function gives an error if a fixed effect term is also included in the random effects, for instance as a grouping level (see the `reg` term in this model). This might not make scientific sense in this particular case, but might be useful for other real-world models.
``` r
library(lme4)
library(insight)

model = lmer(total.fruits ~ amd + reg + (1 | reg:popu), data = Arabidopsis)

pred = get_predicted(model)
#> Error in X %*% fixef(object): non-conformable arguments
```

Second, the terms describing lower groupings (for instance the `reg:popu` term in this model) are not recognized as individual terms, so the function (*silently*, if include_random is "default") drops the random effects from the predictions.
``` r
model2 = lmer(total.fruits ~ amd + (1 | reg:popu), data = Arabidopsis)

pred2 = get_predicted(model2, include_random = TRUE)
#> Warning: `include_random` was set to `TRUE`, but not all random effects were
#>   found in `data`.
#>   Setting `include_random = FALSE` now.

packageVersion("lme4")
#> [1] '1.1.35.5'
packageVersion("insight")
#> [1] '0.20.4'
```

<sup>Created on 2024-12-16 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

strengejacke:
Fixes #981


DominiqueMakowski:
While you're at it, note also the get_predicted() issue which might be related to datagrid

https://github.com/easystats/modelbased/issues/288#issuecomment-2590526203

PR #1024, opened 2025-03-21
`null_model()` only runs on complete data (for correct logLik of NullModel)

strengejacke:


bwiernik:
Just asking, don't we have a function that already returns the exact model frame that was used for fitting, post dropping missing cases, etc? Can we use that for robustness (eg, modeling functions that impute missingness)?

strengejacke:
We do. `get_data()` has a `source` argument, to access data in the environment, or from the model frame. But the model frame has term names as column names, leading to errors:

``` r
d <- data.frame(y = rnorm(100), x = rnorm(100))
m <- lm(y / 10 ~ x, data = d)

mf <- insight::get_data(m, source = "mf")
head(mf)
#>          y/10           x
#> 1  0.19387507  0.16238172
#> 2  0.03514922 -0.09847932
#> 3  0.14609369 -0.62099325
#> 4  0.02314441 -0.58306030
#> 5  0.05217063 -0.29653352
#> 6 -0.01462033  0.16495938

update(m, ~1, data = mf)
#> Error in eval(predvars, data, env): object 'y' not found
```

<sup>Created on 2025-03-21 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

PR #1039, opened 2025-04-07
Better handle decision models (RT + Choice)

### LINKED ISSUE #1038: Make `get_predicted()` work with Wiener models from brms
This is a follow-up to add special treatment for some more families, by adding a `Response` column:

### wiener

[Wiener models](https://github.com/DominiqueMakowski/cogmod?tab=readme-ov-file#drift-diffusion-model-ddm) are heavily used in cognitive science to model reaction time and binary choice at the same time. 

``` r
m_ddm <- readRDS(url("https://raw.github.com/DominiqueMakowski/cogmod/main/man/figures/m_ddm.rds"))
# m_ddm

insight::get_predicted(m_ddm, iterations = 5, data = insight::get_data(m_ddm)[1:4,]) |> 
  as.data.frame()
#>   Predicted    iter_1    iter_2    iter_3    iter_4    iter_5
#> 1 0.4741491 0.4711659 0.4800093 0.4692831 0.4753183 0.4749688
#> 2 0.4741491 0.4711659 0.4800093 0.4692831 0.4753183 0.4749688
#> 3 0.4741491 0.4711659 0.4800093 0.4692831 0.4753183 0.4749688
#> 4 0.4741491 0.4711659 0.4800093 0.4692831 0.4753183 0.4749688

brms::posterior_predict(m_ddm, ndraws=5, newdata = insight::get_data(m_ddm)[1:4,], negative_rt = TRUE)
#>            [,1]       [,2]       [,3]       [,4]
#> [1,] -0.5142505  0.4978713 -0.2910381  1.0126201
#> [2,] -0.4426298 -0.4950428 -0.3088033 -0.3641679
#> [3,]  0.5410477  0.4704078  0.3633962 -0.2897229
#> [4,]  0.5705412  0.4603269 -0.3960405 -0.2964604
#> [5,]  0.7273362 -0.6391990 -0.2595111 -0.2871427
```

<sup>Created on 2025-04-05 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

Currrently the predictions are well pulled. `wiener` models are treated specially in brms, for which there is the unique argument `negative_rt` that should be set to true when pulling the predictions, which provides an indication about the choice. 

We should do that for this family,  and then reformat by:
- Negative RTs = `Response` = 1, positive = `Response` = 0 *(note: could be the opposite, not sure of the mapping)*
- Re-positive the negative RTs, but add a `Response` column (with zero or one)

For wiener, the number of observations should be the same as the data (unlike for `categorical` that returns a value for *each* response, here the response is an outcome per draw as well) 

 _Originally posted by @DominiqueMakowski in [#1034](https://github.com/easystats/insight/issues/1034#issuecomment-2780698588)_

DominiqueMakowski:
Fix #1038 

It *essentially* works:

``` r
x <- readRDS(url("https://raw.github.com/DominiqueMakowski/cogmod/main/man/figures/m_ddm.rds"))
# x <- insight::download_model("brms_categorical_1_fct")
# x <- insight::download_model("brms_mv_1")

data <- insight::get_data(x)[1:5, ]

# Useful initialization of stuff to interactively dig into get_predicted
# predict = "prediction"; iterations = 10; ci = NULL; ci_method = "quantile"; include_random = "default"; include_smooth = TRUE; verbose = TRUE; dots <- list()
# my_args <- insight:::.get_predicted_args(
#   x,
#   data = data,
#   predict = predict,
#   include_random = include_random,
#   include_smooth = include_smooth,
#   ci = ci,
#   verbose = verbose
# )


insight::get_predicted(x, data = data, predict = "prediction", iterations = 3)
#> Predicted values:
#> 
#>    Row Response        rt response Predicted
#> 1    1       rt 0.9567780        1 1.0434566
#> 2    2       rt 0.6907724        0 0.7045342
#> 3    3       rt 1.1837892        0 1.0207763
#> 4    4       rt 1.5447240        0 1.5272589
#> 5    5       rt 1.4839641        0 1.0195394
#> 6    1 response 0.9567780        1 0.3333333
#> 7    2 response 0.6907724        0 0.3333333
#> 8    3 response 1.1837892        0 0.3333333
#> 9    4 response 1.5447240        0 0.3333333
#> 10   5 response 1.4839641        0 0.6666667
#> 
#> NOTE: Confidence intervals, if available, are stored as attributes and can be accessed using `as.data.frame()` on this output.
```

<sup>Created on 2025-04-07 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

The idea is essentially to "pretend" that it is a multivariate output by having 3D matrix of draws.

It basically works, although the output is slightly confusing because we have `Response` as a column name for the type of prediction (response vs. rt), `response` which is  the actual observed response, and  then `Predicted` which corresponds to the average of draws for both type of outputs, respectively...

Not sure how to prevent issues if say the data column is named `Response` (and clashes with the `Response` that we add)


DominiqueMakowski:
``` r
library(brms)
library(cogmod)  # remotes::install_github("DominiqueMakowski/cogmod")

x <- readRDS(url("https://raw.github.com/DominiqueMakowski/cogmod/main/man/figures/m_lnr.rds"))

data <- insight::get_data(x)[1:5, ]

insight::get_predicted(x, data = data, predict = "prediction", iterations = 3) |> 
  as.data.frame()
#>    Row Response        rt response Predicted    iter_1    iter_2    iter_3
#> 1    1       rt 0.9567780        1 0.8705663 0.9629671 0.4436473 1.2050844
#> 2    2       rt 0.6907724        0 1.1298689 1.0259591 1.8651514 0.4984964
#> 3    3       rt 1.1837892        0 0.4684940 0.4888795 0.4962847 0.4203177
#> 4    4       rt 1.5447240        0 1.4890139 1.5127369 0.3090426 2.6452620
#> 5    5       rt 1.4839641        0 0.5149100 0.3395105 0.8112314 0.3939880
#> 6    1 response 0.9567780        1 0.3333333 0.0000000 1.0000000 0.0000000
#> 7    2 response 0.6907724        0 0.3333333 1.0000000 0.0000000 0.0000000
#> 8    3 response 1.1837892        0 0.3333333 1.0000000 0.0000000 0.0000000
#> 9    4 response 1.5447240        0 0.6666667 1.0000000 0.0000000 1.0000000
#> 10   5 response 1.4839641        0 0.6666667 1.0000000 0.0000000 1.0000000
```

<sup>Created on 2025-04-07 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>


Decision-making models (a distribution that jointly models RT and response) are a *big* thing in psychology, but its support is dramatically lacking (the main reason I learn Julia was to be able to use them). One of the tricky bit is that since the distribution spits out two values (rt + integer of choice), it doesn't typically play nice with most modeling software or framework, and hence requires the use of bespoke tools.

In Julia, the SequentialSamplingModels package leveraged the power and flexibility of Julia by essentially implementing a new type of multivariate distribution, but it remains a bit awkward because predictions from these models typically contain RT and choice interwoven within the output, which requires bespoke postprocessing to delineate RTs and choice. 

In the only decision-model implemented in brms, the wiener model, the workaround is to have negative RTs, which works because the choice can only be binary. But that doesn't work for other models that can have more than 2 categories of response.

I've **FINALLY** (I've been on it very actively since 4 days non-stop, and passively for a long time) managed to implement another simple model in brms using custom Stan families, the LogNormal Race model. It is quite simple in itself but was a pain to implement due to particularities of Stan and brms. 

Anyway, all that to say that IMO the easiest way to implement these models leads to them spitting, for each observation, 2 predictions (rt + choice), which leads to twice as many "iterations" (we get 10 columns if we ask for 5 iterations, with the first column being "iter1_rt" and the second being "iter1_choice"). 

It is quite straightforward to then delineate them and reorganize them, which is what these new few lines of code do.

It feels a bit weird to add some code to easystats to support custom families (given that it's not yet an "official" well-established brms family), **BUT** the code is quite minimal and I engage myself to provide the maintenance support for that. The upside is that if we make easystats work nicely with decision-making models, it might be a *big* thing (many many potential users...)

DominiqueMakowski:
modelbased::estimate_means() or modelbased::estimate_relation() look like?

Good question, my guess is that it would need two predicted columns, `Probability` (of e.g., choice 1 vs. 0) and `Predicted` (predicted time). It's like a combined logistic + linear model

DominiqueMakowski:
Which, now that I think of it, might mean that it would be more consistent to have this "multiple" prediction framework implemented at get_predicted level, with all iterations Postfixed based on what their index `iter_1_response`, `iter_1_rt`, rather than having them as separate rows? This would also potentially apply to multivariate models

strengejacke:
What about two lists, one with predictions of the probability of response = 0/1, and one with the predictions of the RT (response)?

DominiqueMakowski:
I thought about this during the night, and I wonder if, for `categorical` and `multivariate` and decision making models (wiener, LNR, ...), the easiest is not to:
- Keep doing the same we do
- But rename the `Response` to `.resp` to avoid potential name conflicts (and also `Row` to `.row`)
- At the very end of the function, we check: if we detect the presence of `.row` and `.resp`, it means the predictions are multivariate, and we then "simply" `data_to_wide()` the mean predictions + all iterations based on `.resp`

Benefits:
- Minimal changes to the current pipeline (simply add a reshape step at the end)
- Consistent output (nrows of predictions always matches nrows of data)
- No name conflict (since `.resp` and `.row`  become temporary and can then be removed)
- Nice for mv models because the predictions will then be like `Predicted_mpg`, `Predicted_qsec`, and `iter_1_mpg`, `iter_1_qsec` etc. (assuming mpg and qsec where the multivariate outcomes), and I think it's quite "readable"

Cons:
- Breaking change for `categorical` family
- Need to adapt `reshape_iterations()` to take into account with the response suffix and creating different columns for different responses 
- Needs probably to adapt modelbased postprocessing to also group predictions by suffix
- Note: this reshaping might be computationally expensive (I have models with  960000 iterations for like 1000 parameters 😢 ), but we can try to profile it and make `data_to_wide()` as efficient as possible in the future (assuming it is even worth it, there might a lot of duration overhead coming from posterior_predict in the first place).

wdyt?

strengejacke:
I think there is a difference between models where the outcome has _multiple levels_ (multinomial, polr), or where you have different _types of outcomes_ (wiener, decision stuff). Or am I misunderstanding the decision models?

DominiqueMakowski:
> difference between models

Good question. Mmh, you're right that it somehow makes sense to treat factor levels *as levels* (as we do now) and have them in long format. I guess for multivariate models (I think we can consider decision models to be equivalent to MV models - I think they conceptually are, just brms doesn't label them as such) it's a bit weirder given that the predictions correspond to fundamentally different "stuff" (and different types, like values or probability or integers), so  it's  a bit weirder to have them as different rows in the same column. 

On the other hand, on some level it's more elegant to have a consistent output...

Maybe we could keep things as-is-now here (i.e., "long" format where `nrows(data)` can sometimes `!= nrows(draws)`, but N iter always == `ncols(iter)`, and do the reshaping into multiple columns for MV models at `modelbased` level? 

There's still the issue of name clashing, I often have `Response` in my own data, which I could use inside a decision model, and I think that would break it, so we need maybe to make sure to throw a warning or something if one of the data variable is named `Response`

strengejacke:
Can you clarify a bit what we want to achieve?

- I wonder, what is the _actual output_ from `posterior_epred()` and similar for the Wiener model?
- Which additional information do we get when we set `negative_rt = TRUE`?
- Which information (column) is missing that you manually create? Which is the column with the information that is used for transformation? You're doing something like if x < 0 then 0 else 1, but for which output of `posterior_predictions()`?
- If we create a new column manually, why do we need to multiply iterations as well?

I think I still don't quite understand the structure of the _original_ output from `posterior_predictions()`, and the desired structure of the _modified_ output, and why we need this reshaping...

DominiqueMakowski:
1. posterior epred

``` r
library(brms)
#> Loading required package: Rcpp
#> Loading 'brms' package (version 2.21.5). Useful instructions
#> can be found by typing help('brms'). A more detailed introduction
#> to the package is available through vignette('brms_overview').
#> 
#> Attaching package: 'brms'
#> The following object is masked from 'package:stats':
#> 
#>     ar
```

``` r

x <- readRDS(url("https://raw.github.com/DominiqueMakowski/cogmod/main/man/figures/m_ddm.rds"))

brms::posterior_epred(x)[1:5, 1:5]
#> Loading required package: rstan
#> Loading required package: StanHeaders
#> 
#> rstan version 2.32.6 (Stan version 2.32.2)
#> For execution on a local, multicore CPU with excess RAM we recommend calling
#> options(mc.cores = parallel::detectCores()).
#> To avoid recompilation of unchanged Stan programs, we recommend calling
#> rstan_options(auto_write = TRUE)
#> For within-chain threading using `reduce_sum()` or `map_rect()` Stan functions,
#> change `threads_per_chain` option:
#> rstan_options(threads_per_chain = 1)
#> Do not specify '-march=native' in 'LOCAL_CPPFLAGS' or a Makevars file
#>           [,1]      [,2]      [,3]      [,4]      [,5]
#> [1,] 0.8339147 0.8339147 0.8339147 0.8339147 0.8339147
#> [2,] 0.8618990 0.8618990 0.8618990 0.8618990 0.8618990
#> [3,] 0.8532544 0.8532544 0.8532544 0.8532544 0.8532544
#> [4,] 0.8186870 0.8186870 0.8186870 0.8186870 0.8186870
#> [5,] 0.8421000 0.8421000 0.8421000 0.8421000 0.8421000
```

<sup>Created on 2025-04-08 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

(That said, expected prediction don't make *much* sense for most decision making models as the "mean" of the RT is not easily obtainable. The parameters of the model often are non-distribution-summary variables (mean, median etc.) but instead stuff like evidence accumulation rate etc. So people should mostly use `posterior_predict` than `epred`


2. Setting negative_rt = TRUE allows to extract the information about the response, which is otherwise hidden (as if the wiener was a univariate model that only predicted reaction times)


3 and 4: There is one iteration "per output type", i.e., in the raw output (with negative_rt), each current iteration value combines information from two actual iterations, one about rt and the other about choice

This is why we need to have "more" iteration columns, because the model actually produces more iterations. Or more accurately, it produce two predictions for each iteration

strengejacke:
Has `posterior_predictions()` a similar output? It takes hours to get a result from your example model, so I stopped in between. 😄 

>  each current iteration value combines information from two actual iterations, 

This is probably the most confusing part. Do we have _one_ column with iterations, or _two_ columns? Or do you want to say that there is _one_ value that contains _two_ different information?

DominiqueMakowski:
In brms::posterior_predict, it returns *one* iteration with *two* different information (identifiable via the sign and absolute value)

But in easystats, we should post process that to:
- Either have different *rows* for the different predicted variables (choice and RT). this is what we have now implemented in this PR
- Either have different *cols* for the different variables, i.e., separate each `iter_` into its `_response` and `_rt` part

bwiernik:
We should do 1 response per row (similar to how we do other multivariate models)

strengejacke:
Ok, if we have

> In brms::posterior_predict, it returns one iteration with two different information (identifiable via the sign and absolute value)

Then we have the many number of draws to compute point estimate and uncertainty. But why would we need _additional iteration columns_? We could just compute the point estimate of the many draws for each parameter, and then for each of those point estimates create a new column that is, say, 0 for negative point estimate and 1 for positive.

I still don't get the point of having additional iteration columns, and especially: where do these additional columns come from? Or do you want to _duplicate_ all iterations, and split all duplicated iteration columns at their sign to have new iterations that either are 0 or 1, then take the mean oder median of all those 0 and 1, and then again set the resulting "point estimate" to either 0 or 1, based on the mean of all 0/1 iterations? That sounds like an unnecessary step to me.

DominiqueMakowski:
Then we should still do some reshaping in modelbased, for instance for decision models we could get one column named `Probability` (the probability of making choice 1 vs. 0; as for logistic models) and `Predicted` (for RT; as for linear models)

strengejacke:
Also, why are the two outputs identical?

``` r
x <- readRDS(url("https://raw.github.com/DominiqueMakowski/cogmod/main/man/figures/m_ddm.rds"))

set.seed(123)
out1 <- brms::posterior_epred(x, negative_rt = TRUE)
set.seed(123)
out2 <- brms::posterior_epred(x, negative_rt = FALSE)

identical(out1, out2)
#> [1] TRUE
```

<sup>Created on 2025-04-08 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

DominiqueMakowski:
For the wiener model in particular, yes you can compute point estimates on the negative + positive data, and then extract the choice from the summary directly yes. 

But I think conceptually this is wrong, because there technically *are* two different values for each iteration. So if you ask have 5 iterations for one observation, you would expect 10 values. It's just that brms condenses the two outputs into one, which works for wiener (because choice must be binary), but that doesn't apply to other models (in which you can have N choices, and so the identification via sign becomes impossible). 

In other words, for a linear ballistic accumulator (LBA), you can model at the same time (for each observation), the reaction time, and the choice between A, B, C and D for instance. And the output of each iteration sample would be `c(0.64, "C")` (this is theoretical as there is no LBA implemented in brms to my knowledge, but it might happen soon). 

 > why should we duplicate all iterations

Because one might want to do something with the choice iteration (e.g., compute the probability, or the mode), and something else with the RT iterations (median, compute a CI, etc.). 

Again, there are fundamentally different quantities returned, and should be treated as such

DominiqueMakowski:
> Also, why are the two outputs identical?

I have no idea what `epred` does for wiener, expected predictions don't really apply easily as a concept to these models...

strengejacke:
What you have here sounds more like a "two part model", like zero-inflated model, to me, where it makes sense to split predictions into separate tables. One for reaction times, one for choices (or whatever the response is), because we really have two different components here, not levels belonging to _one_ "variable".

DominiqueMakowski:
> o it's only the Wiener model that puts all the information in one value, but for other models, the predicted "value" can be a vector of values

Correct, in fact it's already the case, e.g. in the LNR model implemented:

``` r
library(brms)
library(cogmod)  # remotes::install_github("DominiqueMakowski/cogmod")

x <- readRDS(url("https://raw.github.com/DominiqueMakowski/cogmod/main/man/figures/m_lnr.rds"))
brms::posterior_predict(x)[1:9, 1:5]

#>            [,1] [,2]      [,3] [,4]      [,5]
#>  [1,] 0.2323749    1 0.4487325    0 0.3630165
#>  [2,] 1.0390661    1 0.7568267    0 1.4664852
#>  [3,] 8.7807571    0 0.3246564    0 1.3702212
#>  [4,] 0.8860498    1 0.9898137    1 0.3228189
#>  [5,] 1.7039727    0 0.5479201    0 1.0653412
#>  [6,] 0.6784356    0 1.8483422    0 1.3009101
#>  [7,] 0.5170161    0 0.4448472    1 0.3660952
#>  [8,] 1.6239607    1 0.3688161    1 1.7070315
#>  [9,] 0.6326186    0 1.4676291    1 0.9261332
```

<sup>Created on 2025-04-08 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

The second and fourth columns are the choice output and the odd cols are the RT.
This particular LNR model is a 2-choice one, so the choice column is also binary, but this could be easily changed to accommodate any number of responses, and so the choice columns could be integers (0, 1, 2, 3, etc)

DominiqueMakowski:
> What you have here sounds more like a "two part model", like zero-inflated model, to me, where it makes sense to split predictions into separate tables.

I'm not sure, it's more a multivariate model where it tries to predict two different variables at the same time, so it makes sense to keep them in the same table, as the choice and the RT for each observation are somewhat linked

I would be happy with, in `get_predicted()`, having output like:

`Predicted_rt, Predicted_response, iter_1_rt, iter_1_response, ... iter_100_rt, iter_100_response` 

Because this is easily implementable (just need to reshape to wide the current output):
- One can easily extract the corresponding columns of  the variables they are interested in getting
- The number of observations matches the original data frame (which is quite important I feel, it's a bit awkward for categorical models where your data 5 rows and you get 15 rows of prediction)


Alternatively, keep-it as-is in get_predicted (multiple observations per iteration, one for each outcome), and do the reshaping in modelbased:
- `categorical` is the same as `mv` in get_predicted (good)
- Possible naming issues `"Response"` is a dangerous column name when a lot of the data of decision stuff contains some "response". But that's easy to address with a warning or such

DominiqueMakowski:
let's refer to this as the long format where each observation receives *N* rows

bwiernik:
This is the epred for the wiener model in brms https://github.com/paul-buerkner/brms/blob/315c7874d6e1b58eb9082e5e07521682f0dc2ca9/R/posterior_epred.R#L439

strengejacke:
ok, then we should move the new code into the last `else` statement, where we check the predict type (i.e. move some line higher)?

strengejacke:
This is the current implementation. Not sure about your final decision, if you're satisfied with this layout, or would you like to change something?

``` r
x <- readRDS(url("https://raw.github.com/DominiqueMakowski/cogmod/main/man/figures/m_ddm.rds"))

data <- insight::get_data(x)[1:5, ]

insight::get_predicted(x, data = data, iterations = 3)
#> Warning: Using `predict = "expectation"` will return the value of the main model
#>   parameter. If you want predicted reaction times and responses, use
#>   `predict = "prediction"`.
#> Predicted values:
#> 
#> [1] 0.8373492 0.8373492 0.8373492 0.8373492 0.8373492
#> 
#> NOTE: Confidence intervals, if available, are stored as attributes and can be accessed using `as.data.frame()` on this output.

insight::get_predicted(x, data = data, predict = "prediction", iterations = 3)
#> Predicted values:
#> 
#>    Row Response        rt response Predicted
#> 1    1       rt 0.9567780        1 1.0264451
#> 2    2       rt 0.6907724        0 0.6355699
#> 3    3       rt 1.1837892        0 1.7586953
#> 4    4       rt 1.5447240        0 1.3494908
#> 5    5       rt 1.4839641        0 0.8575846
#> 6    1 response 0.9567780        1 0.6666667
#> 7    2 response 0.6907724        0 0.0000000
#> 8    3 response 1.1837892        0 0.6666667
#> 9    4 response 1.5447240        0 0.3333333
#> 10   5 response 1.4839641        0 0.0000000
#> 
#> NOTE: Confidence intervals, if available, are stored as attributes and can be accessed using `as.data.frame()` on this output.


insight::get_predicted(x, data = data, iterations = 3) |> as.data.frame()
#> Warning: Using `predict = "expectation"` will return the value of the main model
#>   parameter. If you want predicted reaction times and responses, use
#>   `predict = "prediction"`.
#>   Predicted    iter_1   iter_2    iter_3
#> 1 0.8450876 0.8498802 0.839725 0.8456576
#> 2 0.8450876 0.8498802 0.839725 0.8456576
#> 3 0.8450876 0.8498802 0.839725 0.8456576
#> 4 0.8450876 0.8498802 0.839725 0.8456576
#> 5 0.8450876 0.8498802 0.839725 0.8456576

insight::get_predicted(x, data = data, predict = "prediction", iterations = 3) |> as.data.frame()
#>    Row Response        rt response Predicted    iter_1    iter_2    iter_3
#> 1    1       rt 0.9567780        1 1.6381931 1.3378063 2.3125856 1.2641875
#> 2    2       rt 0.6907724        0 0.8991299 0.7147916 0.2756385 1.7069597
#> 3    3       rt 1.1837892        0 1.6470780 1.1161759 3.2213941 0.6036640
#> 4    4       rt 1.5447240        0 1.1991846 0.7772626 0.4451159 2.3751753
#> 5    5       rt 1.4839641        0 1.7390333 1.1707719 3.2540836 0.7922445
#> 6    1 response 0.9567780        1 0.3333333 1.0000000 0.0000000 0.0000000
#> 7    2 response 0.6907724        0 0.3333333 0.0000000 0.0000000 1.0000000
#> 8    3 response 1.1837892        0 0.6666667 0.0000000 1.0000000 1.0000000
#> 9    4 response 1.5447240        0 0.3333333 1.0000000 0.0000000 0.0000000
#> 10   5 response 1.4839641        0 0.6666667 0.0000000 1.0000000 1.0000000
```

<sup>Created on 2025-04-08 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

DominiqueMakowski:
Actually forgot what i said i was being dumb, it's best to remove the warning about predict="expectation", it is very useful to be able to predict the different parameters 🙈

DominiqueMakowski:
just that:

``` r
x <- readRDS(url("https://raw.github.com/DominiqueMakowski/cogmod/main/man/figures/m_ddm.rds"))

data <- insight::get_data(x)[1:5, ]

pred <- insight::get_predicted(x, data = data, predict = "prediction", iterations = 3) |> 
  as.data.frame()


datawizard::data_to_wide(pred, 
                         values_from = c("Predicted", paste0("iter_", 1:3)),
                         names_from = "Response")
#>   Row        rt response iter_1_response iter_1_rt iter_2_response iter_2_rt
#> 1   1 0.9567780        1               0 0.9900690               0 0.5837879
#> 2   2 0.6907724        0               0 0.9086772               1 1.7099515
#> 3   3 1.1837892        0               0 0.3159253               0 2.1760707
#> 4   4 1.5447240        0               0 1.2147999               0 1.5888926
#> 5   5 1.4839641        0               0 3.5146502               0 1.6269853
#>   iter_3_response iter_3_rt Predicted_response Predicted_rt
#> 1               1 1.0974973          0.3333333    0.8904514
#> 2               1 1.3897503          0.6666667    1.3361263
#> 3               0 0.8719316          0.0000000    1.1213092
#> 4               0 0.3881527          0.0000000    1.0639484
#> 5               0 2.7975203          0.0000000    2.6463852
```

<sup>Created on 2025-04-08 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

DominiqueMakowski:
We would just need to reorder to put the `Predicted_*` back before the `iter_*` but that's essentially it

strengejacke:
And the summarized predictions, like we show them in modelbased? Two predicted, CI and SE columns?

bwiernik:
I think it's very confusing to have multiple sets of response columns. I think it makes more sense to be more consistent with other multivariate models and have two sets of rows. It is inconsistent behavior and means that code that works for one model won't work for another. 

What's the benefit of having multiple sets of columns vs different response groups in different rows?

strengejacke:
I think Dom would like to `cbind()` the original data to the output. But else, I can't think of any other benefit right now... ?

strengejacke:
This is what we have now. If that's ok, we can merge this PR:

``` r
library(brms)
x <- readRDS(url("https://raw.github.com/DominiqueMakowski/cogmod/main/man/figures/m_ddm.rds"))

data <- insight::get_data(x)[1:5, ]

insight::get_predicted(x, data = data, predict = "prediction", iterations = 3) |> 
  as.data.frame()
#>    Row Response        rt response Predicted    iter_1    iter_2    iter_3
#> 1    1       rt 0.9567780        1 0.5705442 0.9505106 0.2714929 0.4896290
#> 2    2       rt 0.6907724        0 0.8962102 1.0506256 0.8112120 0.8267931
#> 3    3       rt 1.1837892        0 2.1708007 2.8496458 1.1017146 2.5610418
#> 4    4       rt 1.5447240        0 0.8051161 0.8372586 1.1796060 0.3984837
#> 5    5       rt 1.4839641        0 0.6631038 0.3938741 0.9018062 0.6936311
#> 6    1 response 0.9567780        1 0.6666667 1.0000000 0.0000000 1.0000000
#> 7    2 response 0.6907724        0 0.3333333 0.0000000 0.0000000 1.0000000
#> 8    3 response 1.1837892        0 0.3333333 0.0000000 0.0000000 1.0000000
#> 9    4 response 1.5447240        0 0.6666667 1.0000000 1.0000000 0.0000000
#> 10   5 response 1.4839641        0 0.0000000 0.0000000 0.0000000 0.0000000

modelbased::estimate_prediction(x, data = data)
#> Model-based Predictions
#> 
#> Response |   rt | response | Predicted |   SE |       95% CI
#> ------------------------------------------------------------
#> rt       | 0.96 |     1.00 |      1.20 | 0.87 | [0.32, 3.56]
#> rt       | 0.69 |     0.00 |      1.15 | 0.78 | [0.33, 3.27]
#> rt       | 1.18 |     0.00 |      1.17 | 0.83 | [0.33, 3.39]
#> rt       | 1.54 |     0.00 |      1.18 | 0.82 | [0.32, 3.24]
#> rt       | 1.48 |     0.00 |      1.22 | 0.86 | [0.33, 3.45]
#> response | 0.96 |     1.00 |      0.40 | 0.49 | [0.00, 1.00]
#> response | 0.69 |     0.00 |      0.43 | 0.49 | [0.00, 1.00]
#> response | 1.18 |     0.00 |      0.40 | 0.49 | [0.00, 1.00]
#> response | 1.54 |     0.00 |      0.41 | 0.49 | [0.00, 1.00]
#> response | 1.48 |     0.00 |      0.45 | 0.50 | [0.00, 1.00]
#> 
#> Variable predicted: rt
#> Predictions are on the response-scale.
```

<sup>Created on 2025-04-09 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

DominiqueMakowski:
> What's the benefit


- Major: Consistent output shape. With the wide format the predictions are always of the same *nrow* as the data. This makes it easily cbindable, but also is less surprising. With the long format your nrows is variable based on the number of levels / outcomes, which makes it harder to use robustly and programmatically. I fully agree with your comment on consistency, but that's the thing currently the output of get_predicted is *not* consistent between most models and `categorical` (and `mv`). By having a wide format (potentially for all of these cases), we enforce length consistency. 
  - You could argue that `ncol` becomes then inconsistent, but given that  the `ncol` is already something that can arbitrarily change (based on how many iterations people extract), I don't think it's a big issue to have iterations for each outcome separately
  - We already have a framework for handling e.g., multiple `CI` (for different alphas), but yes one thing to consider is how our other functions will play with suffixed iterations (i.e., "grouped" iterations). 
    - Note that the problem of correctly summarizing might also become tricky for the long format, e.g., for multivariate models that generate different types of outcomes (e.g., probability & 0-1-2-3 integers) for which differentiated summaries might be needed
- Conceptual: In all of these models, including `categorical`, the model is not making predictions at various levels, it is actually generating more than one prediction at the same time *per iteration*, and these predictions are independent things, so having in a long format doesn't reflect what's truly going on conceptually
  - This point applies arguably less to `categorical`, which is kind of in the middle and for which arguments for both sides would apply 
- Minor: the wide format prevents the potential name clashes due to the `Response` column (quite minor but particularly relevant for decision models where one of the outcomes is most of the time named `"response"`)

All-in-all, it is not a big issue, I'm okay to go with the long format. I was trying to think of what's the most useful, accurate and future-proof approach, but the long format at least "easy" to deal with and doesn't introduce new concepts into easystats (suffixed iterations/summaries), which makes it less dangerous. And doing the reshaping is literally a one-liner 🤷

DominiqueMakowski:
We still need to think about that for modelbased, where we name the predictions based on the predicted type `Predicted`, `Probability`, `Proportion`, but that doesn't apply to predictions with multiple types as we have with the long format, so having the wide format there is better I think (especially since modelbased integrates the original data by default, so there the long format would be less appropriate)

DominiqueMakowski:
> And the summarized predictions, like we show them in modelbased? Two predicted, CI and SE columns?

Yes, in modelbased, I'd do with suffixed summaries. And by default have `reshape_iteration()` also regroup by outcome ("back" to the long format) - we could add an argument `by_outcome = TRUE` to toggle whether reshape_iterations re-groups by outcome too or not (and gets suffixed `iter_value_rt`, `iter_value_response`)

strengejacke:
For "decision models", the wide format might indeed have the advantage that you can have predictions at different scales (though I still think separate tables like for zero-inflated models is better, but that's just a matter of output - even those models are still saved/returned in "long" format by `model_parameters()`!).

But for categorical/polr/multinomial/... you indeed get one prediction per observation and response level, and every "effects" package I know (there are probably not many in R, namely _effects_, _emmeans_ and _marginaleffects_, if I name those not developed by us) displays the results in long format. I don't think anything else is the better choice for displaying predictions of those models.

It seems that you are particularly looking at the Bayesian case, and that you want to combine a) summarized predictions, b) posterior draws, and c) the raw / original data. Therefore, you may need the wide format. But we usually store multiple components in long format, add a `Component` column, and then split by components in the output. We do this for every model with several components or effects, like zero-inflated, random effects, and so on. And my impression is, we should do the same here for the "decision models", which just have multiple components.

We literally _nowhere_ duplicate columns to show "wide" tables - except `compare_parameters()`, where we put models side-by-side.

strengejacke:
See here: the column in the data frame is named `Coefficient`, but we have several tables, where the columns for the coefficients have different names:

``` r
data("fish", package = "insight")
m1 <- glmmTMB::glmmTMB(
  count ~ child + camper + (1 | persons),
  ziformula = ~ child + camper + (1 | persons),
  data = fish,
  family = glmmTMB::truncated_poisson()
)
out <- parameters::model_parameters(m1)

as.data.frame(out)
#>        Parameter Coefficient         SE   CI     CI_low    CI_high           z
#> 1    (Intercept)   1.2628034 0.47558553 0.95  0.3306729  2.1949339   2.6552603
#> 2          child  -1.1416515 0.09304893 0.95 -1.3240240 -0.9592789 -12.2693676
#> 3        camper1   0.7335393 0.09345517 0.95  0.5503705  0.9167080   7.8491031
#> 4    (Intercept)  -0.3893858 0.65229468 0.95 -1.6678599  0.8890883  -0.5969477
#> 5          child   2.0540669 0.30990003 0.95  1.4466740  2.6614598   6.6281597
#> 6        camper1  -1.0082269 0.32324236 0.95 -1.6417703 -0.3746835  -3.1191051
#> 7 SD (Intercept)   0.9311967         NA 0.95  0.4586206  1.8907292          NA
#> 8 SD (Intercept)   1.1739923         NA 0.95  0.5369964  2.5666051          NA
#>   df_error            p Effects   Group     Component
#> 1      Inf 7.924723e-03   fixed           conditional
#> 2      Inf 1.322825e-34   fixed           conditional
#> 3      Inf 4.190230e-15   fixed           conditional
#> 4      Inf 5.505423e-01   fixed         zero_inflated
#> 5      Inf 3.398975e-11   fixed         zero_inflated
#> 6      Inf 1.814012e-03   fixed         zero_inflated
#> 7       NA           NA  random persons   conditional
#> 8       NA           NA  random persons zero_inflated

out
#> # Fixed Effects (Count Model)
#> 
#> Parameter   | Log-Mean |   SE |         95% CI |      z |      p
#> ----------------------------------------------------------------
#> (Intercept) |     1.26 | 0.48 | [ 0.33,  2.19] |   2.66 | 0.008 
#> child       |    -1.14 | 0.09 | [-1.32, -0.96] | -12.27 | < .001
#> camper [1]  |     0.73 | 0.09 | [ 0.55,  0.92] |   7.85 | < .001
#> 
#> # Fixed Effects (Zero-Inflation Component)
#> 
#> Parameter   | Log-Odds |   SE |         95% CI |     z |      p
#> ---------------------------------------------------------------
#> (Intercept) |    -0.39 | 0.65 | [-1.67,  0.89] | -0.60 | 0.551 
#> child       |     2.05 | 0.31 | [ 1.45,  2.66] |  6.63 | < .001
#> camper [1]  |    -1.01 | 0.32 | [-1.64, -0.37] | -3.12 | 0.002 
#> 
#> # Random Effects Variances
#> 
#> Parameter               | Coefficient |       95% CI
#> ----------------------------------------------------
#> SD (Intercept: persons) |        0.93 | [0.46, 1.89]
#> 
#> # Random Effects (Zero-Inflation Component)
#> 
#> Parameter               | Coefficient |       95% CI
#> ----------------------------------------------------
#> SD (Intercept: persons) |        1.17 | [0.54, 2.57]
#> 
#> Uncertainty intervals (equal-tailed) and p-values (two-tailed) computed
#>   using a Wald z-distribution approximation.
#> 
#> The model has a log- or logit-link. Consider using `exponentiate =
#>   TRUE` to interpret coefficients as ratios.
```

<sup>Created on 2025-04-09 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

DominiqueMakowski:
I think these are all good arguments, long format it is! Let's just add a test to make sure that if the model is `rt | dec(Response)` it doesn't create some issues of name clash, but yeah otherwise lgtm

strengejacke:
WDYT? Here's a (quick) suggestion how to handle the Wiener models:

``` r
x <- readRDS(url("https://raw.github.com/DominiqueMakowski/cogmod/main/man/figures/m_ddm.rds"))
data <- insight::get_data(x)[1:5, ]
insight::get_predicted(x, data = data, predict = "prediction", iterations = 3)
#> Predicted values:
#> 
#>    Row Component        rt response Predicted
#> 1    1        rt 0.9567780        1 1.1484224
#> 2    2        rt 0.6907724        0 0.9296970
#> 3    3        rt 1.1837892        0 1.3480989
#> 4    4        rt 1.5447240        0 0.8949522
#> 5    5        rt 1.4839641        0 0.8654259
#> 6    1  response 0.9567780        1 0.6666667
#> 7    2  response 0.6907724        0 0.6666667
#> 8    3  response 1.1837892        0 0.3333333
#> 9    4  response 1.5447240        0 0.6666667
#> 10   5  response 1.4839641        0 0.3333333
#> 
#> NOTE: Confidence intervals, if available, are stored as attributes and can be accessed using `as.data.frame()` on this output.

out <- modelbased::estimate_prediction(x, data = data)
out <- split(out, out$Component)
lapply(names(out), function(i) {
  out[[i]]$Component <- NULL
  out[[i]][[setdiff(names(out), i)]] <- NULL
  tt <- attributes(out[[i]])$table_title
  attr(out[[i]], "table_title") <- paste(tt, ":", i)

  # must be more flexible
  if (i == "response") {
    attr(out[[i]], "table_footer") <- NULL
  }

  out[[i]]
})
#> [[1]]
#> Model-based Predictions : response
#> 
#> response | Predicted |   SE |       95% CI
#> ------------------------------------------
#> 1.00     |      0.41 | 0.49 | [0.00, 1.00]
#> 0.00     |      0.42 | 0.49 | [0.00, 1.00]
#> 0.00     |      0.40 | 0.49 | [0.00, 1.00]
#> 0.00     |      0.43 | 0.50 | [0.00, 1.00]
#> 0.00     |      0.40 | 0.49 | [0.00, 1.00]
#> 
#> [[2]]
#> Model-based Predictions : rt
#> 
#> rt   | Predicted |   SE |       95% CI
#> --------------------------------------
#> 0.96 |      1.18 | 0.82 | [0.33, 3.34]
#> 0.69 |      1.19 | 0.82 | [0.35, 3.22]
#> 1.18 |      1.20 | 0.85 | [0.33, 3.52]
#> 1.54 |      1.18 | 0.81 | [0.34, 3.35]
#> 1.48 |      1.17 | 0.87 | [0.33, 3.46]
#> 
#> Variable predicted: rt
#> Predictions are on the response-scale.
```

<sup>Created on 2025-04-09 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

strengejacke:
Also with `keep_iterations`:

``` r
x <- readRDS(url("https://raw.github.com/DominiqueMakowski/cogmod/main/man/figures/m_ddm.rds"))
data <- insight::get_data(x)[1:5, ]
insight::get_predicted(x, data = data, predict = "prediction", iterations = 3)
#> Predicted values:
#> 
#>    Row Component        rt response Predicted
#> 1    1        rt 0.9567780        1 0.8964064
#> 2    2        rt 0.6907724        0 1.8186247
#> 3    3        rt 1.1837892        0 1.6042676
#> 4    4        rt 1.5447240        0 1.4243915
#> 5    5        rt 1.4839641        0 1.1993594
#> 6    1  response 0.9567780        1 0.6666667
#> 7    2  response 0.6907724        0 0.6666667
#> 8    3  response 1.1837892        0 0.3333333
#> 9    4  response 1.5447240        0 0.6666667
#> 10   5  response 1.4839641        0 0.6666667
#> 
#> NOTE: Confidence intervals, if available, are stored as attributes and can be accessed using `as.data.frame()` on this output.

out <- modelbased::estimate_prediction(x, data = data, keep_iterations = 5)
out <- split(out, out$Component)
lapply(names(out), function(i) {
  out[[i]]$Component <- NULL
  out[[i]][[setdiff(names(out), i)]] <- NULL
  tt <- attributes(out[[i]])$table_title
  attr(out[[i]], "table_title") <- paste(tt, ":", i)

  # must be more flexible
  if (i == "response") {
    attr(out[[i]], "table_footer") <- NULL
  }

  out[[i]]
})
#> [[1]]
#> Model-based Predictions : response
#> 
#> response | Predicted |   SE |       95% CI | iter_1 | iter_2 | iter_3 | iter_4 | iter_5
#> ---------------------------------------------------------------------------------------
#> 1.00     |      0.39 | 0.49 | [0.00, 1.00] |   0.00 |   1.00 |   0.00 |   0.00 |   1.00
#> 0.00     |      0.41 | 0.49 | [0.00, 1.00] |   0.00 |   0.00 |   0.00 |   1.00 |   0.00
#> 0.00     |      0.41 | 0.49 | [0.00, 1.00] |   0.00 |   0.00 |   1.00 |   1.00 |   1.00
#> 0.00     |      0.41 | 0.49 | [0.00, 1.00] |   1.00 |   0.00 |   0.00 |   0.00 |   0.00
#> 0.00     |      0.40 | 0.49 | [0.00, 1.00] |   0.00 |   0.00 |   0.00 |   0.00 |   1.00
#> 
#> [[2]]
#> Model-based Predictions : rt
#> 
#> rt   | Predicted |   SE |       95% CI | iter_1 | iter_2 | iter_3 | iter_4 | iter_5
#> -----------------------------------------------------------------------------------
#> 0.96 |      1.17 | 0.80 | [0.32, 3.19] |   0.85 |   3.61 |   1.89 |   2.85 |   0.90
#> 0.69 |      1.21 | 0.90 | [0.32, 3.57] |   1.04 |   0.51 |   1.48 |   1.92 |   0.47
#> 1.18 |      1.16 | 0.78 | [0.30, 3.17] |   0.52 |   0.56 |   0.77 |   0.70 |   1.05
#> 1.54 |      1.14 | 0.77 | [0.32, 3.20] |   1.78 |   0.44 |   0.38 |   0.36 |   1.05
#> 1.48 |      1.14 | 0.78 | [0.33, 3.21] |   0.34 |   0.40 |   0.42 |   0.76 |   1.28
#> 
#> Variable predicted: rt
#> Predictions are on the response-scale.
```

<sup>Created on 2025-04-09 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

DominiqueMakowski:
you mean how to handle them in modelbased?

strengejacke:
Yes.

DominiqueMakowski:
I'm not a big fan of the multiple table printing to be entirely fair 😁 (I like when the printing is somewhat related to what's under, in this case one table), but yeah I think in this case it makes sense.

And indeed, users can to the wide-reshaping themselves

strengejacke:
<sup>(Dom completely desperate now...)</sup>

Or like this:

``` r
x <- readRDS(url("https://raw.github.com/DominiqueMakowski/cogmod/main/man/figures/m_ddm.rds"))
data <- insight::get_data(x)[1:5, ]
insight::get_predicted(x, data = data, predict = "prediction", iterations = 3)
#> Predicted values:
#> 
#>    Row Component        rt response Predicted
#> 1    1        rt 0.9567780        1 1.3058946
#> 2    2        rt 0.6907724        0 1.3471514
#> 3    3        rt 1.1837892        0 0.4506750
#> 4    4        rt 1.5447240        0 0.6872541
#> 5    5        rt 1.4839641        0 0.9263969
#> 6    1  response 0.9567780        1 0.3333333
#> 7    2  response 0.6907724        0 0.6666667
#> 8    3  response 1.1837892        0 0.0000000
#> 9    4  response 1.5447240        0 0.0000000
#> 10   5  response 1.4839641        0 0.3333333
#> 
#> NOTE: Confidence intervals, if available, are stored as attributes and can be accessed using `as.data.frame()` on this output.

out <- modelbased::estimate_prediction(x, data = data)
out <- split(out, out$Component)
do.call(cbind, lapply(names(out), function(i) {
  out[[i]]$Component <- NULL
  out[[i]][[setdiff(names(out), i)]] <- NULL
  colnames(out[[i]]) <- paste0(i, "_", colnames(out[[i]]))
  out[[i]]
})) |>
  insight::format_table() |>
  insight::export_table(table_width = Inf)
#> response_response | response_Predicted | response_SE | response  CI | rt_rt | rt_Predicted | rt_SE |       rt  CI
#> -----------------------------------------------------------------------------------------------------------------
#> 1.00              |               0.40 |        0.49 | [0.00, 1.00] |  0.96 |         1.15 |  0.79 | [0.32, 3.25]
#> 0.00              |               0.44 |        0.50 | [0.00, 1.00] |  0.69 |         1.17 |  0.77 | [0.33, 3.18]
#> 0.00              |               0.44 |        0.50 | [0.00, 1.00] |  1.18 |         1.21 |  0.87 | [0.32, 3.49]
#> 0.00              |               0.42 |        0.49 | [0.00, 1.00] |  1.54 |         1.17 |  0.82 | [0.33, 3.43]
#> 0.00              |               0.42 |        0.49 | [0.00, 1.00] |  1.48 |         1.15 |  0.85 | [0.32, 3.39]
```

<sup>Created on 2025-04-09 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>


It _would_ in principal work for Wiener models, since we have _always and only_ two components (right?).

DominiqueMakowski:
Haha well if we are not reshape_wide, I would "simply" suffix the name of what is predicted to all the columns (Predicted + iter), so that it works nicely for all mv models with any number of outcomes

But I think I'm now fine to just leave it in the long-format, y'all convinced me :D

PR #1041, opened 2025-04-09
Deal with lnr models

strengejacke:


R/get_predicted.R
```diff
@@ -945,7 +946,7 @@ get_predicted.phylolm <- function(x,
       stringsAsFactors = FALSE
     )
     # make sure we have the correct name for the "Response column"
-    if (is_wiener) {
+    if (is_wiener || is_lnr) {
```
DominiqueMakowski, line 949: I think we could have something more generic than is_lnr, like `is_ssm` ("sequential sampling models", the official name of decision making models)

DominiqueMakowski: or `is_rtchoice`

strengejacke: We could also change `is_wiener` to the general name

R/utils_model_info.R
```diff
@@ -419,6 +420,7 @@
     is_ftest = is_ftest,
     is_meta = is_meta,
     is_wiener = inherits(x, "brmsfit") && fitfam == "wiener",
+    is_lnr = inherits(x, "brmsfit") && fitfam == "custom" && identical(brms_custom_name, "lnr"),
```
DominiqueMakowski, line 423: here we can have `is_rtchoice`, and we will check for various models (LNR, LBA, ...) that will all behave the same (by having even-odd iterations)
Wiener behaves differently (negative RTs) so we need to deal separately

PR #1047, opened 2025-04-11
modelbased::estimate_grouplevel() misses dpars from ZOIB

### LINKED ISSUE #1046: modelbased::estimate_grouplevel() misses dpars from ZOIB
I'm putting it in insight as it might probably come from here:

``` r
library(brms)

df <- data.frame(
  y = cogmod::rbext(1000),
  participant = rep(paste0("S", 1:10), each = 100)
)

f <- bf(
  y ~ 1 + (1 | participant),
  phi ~ 1 + (1 | participant),
  zoi ~ 1 + (1 | participant),
  coi ~ 1 + (1 | participant)
)

m <- brm(f, data = df, family = zero_one_inflated_beta(), init = 0,
          chains = 4, iter = 500, backend = "cmdstanr", refresh = 0)


g <- modelbased::estimate_grouplevel(m)
plot(g)
```

``` r
unique(g$Component)
#> [1] "conditional" "phi"
```

<sup>Created on 2025-04-11 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

Image

strengejacke:
Fixes #1046

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Type: Package
 Package: insight
 Title: Easy Access to Model Information for Various Model Objects
-Version: 1.1.0.30
+Version: 1.1.0.31
 Authors@R:
     c(person(given = "Daniel",
              family = "Lüdecke",
--- a/R/clean_parameters.R
+++ b/R/clean_parameters.R
@@ -431,7 +431,7 @@ clean_parameters.mlm <- function(x, ...) {
     # we must start with "conditional one", so it's not confused with "conditional"
     com <- if (startsWith(i, "conditional_one_inflated")) {
       "conditional_one_inflated"
-    } else if (grepl("conditional_one_inflated", i, fixed = TRUE) || i == "random") {
+    } else if (grepl("conditional", i, fixed = TRUE) || i == "random") {
       "conditional"
     } else if (grepl("zero_inflated", i, fixed = TRUE)) {
       "zero_inflated"
--- a/R/clean_parameters.R
+++ b/R/clean_parameters.R
@@ -428,7 +428,10 @@ clean_parameters.mlm <- function(x, ...) {
       "fixed"
     }
 
-    com <- if (grepl("conditional", i, fixed = TRUE) || i == "random") {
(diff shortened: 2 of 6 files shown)


strengejacke:
@DominiqueMakowski 

``` r
model <- insight::download_model("brms_zoib_1")
g <- modelbased::estimate_grouplevel(model)
plot(g)
```



<sup>Created on 2025-04-11 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

DominiqueMakowski:
can we retain `coi` and `zoi` by default rather than renaming them?

DominiqueMakowski:
what's the reason for the renaming in this particular case?

strengejacke:
We have these renaming since a very long time. It started with the "family-like" components, e.g. zero-inflated models, so we have `"zero_inflation"` by default for all models. That's why "zi" is "zero_inflated". We did the same for "zoi" and "coi", since these are somewhat "related".

strengejacke:
We could simplify this and use "zoi" and "coi", this would also affect `find_formula()` or `find_parameters()`.

What about `clean_names()`? The components should be named "coi" and "zoi", too?

DominiqueMakowski:
The overarching question is whether we start considering - at least for brms - zero-inflated-related components as regular distributional parameters, rather than as something else.

The pros is that it (probably?) makes things simpler, and consistent. Moreover, I don't think there is a clear cut between ZI-related components and dpars, zi-components *are* just additional dpars (and in some case, like CHOCO/Bext/Ordered beta/Xbeta, the dpars are related to the presence of zero-ones, but not necessarily their "inflation", so the distinction between what is ZI-related and not is murky)

The cons is that it might lead to some minor discrepancies with freq models. 
I never use Frequentist ZI models, so I can't form a good opinion here. But as far as brms goes, I would tend to be in favour of a simplification/unification and just treat zi-related dpars as we treat other dpars. wdyt?

strengejacke:
It would definitely simplify handling of brms models. Since the "dpars" thinking is not that relevant in freq world (where we don't have all these distributional parameters), we might think about just having special brms treatment here.

But that would be something for a new PR.

PR #1058, opened 2025-04-19
Allow dpars-argument to be passed to predict-function

strengejacke:
if user provided `dpars` argument, we just pass it to the `predict()`-function. we don't overwrite the `predict` argument, because this allows us to pass dpars to `posterior_predict()` - else, when dpars are provided via the `predict`-argument, we always set `predict = "expectation"`, thus, have no option to use dpars with `posterior_predictions()`.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Type: Package
 Package: insight
 Title: Easy Access to Model Information for Various Model Objects
-Version: 1.1.0.37
+Version: 1.1.0.38
 Authors@R:
     c(person(given = "Daniel",
              family = "Lüdecke",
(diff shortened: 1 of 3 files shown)


strengejacke:
@DominiqueMakowski can you install the latest main from GitHub and check if all your edge cases work as expected?

DominiqueMakowski:
It should work by default as previously right? So what is the feature allowed by this changed?

strengejacke:
Yes, it's only an addition. We can now also pass the `dpar` argument and use `predict = "prediction"`.

strengejacke:
Like `get_predicted(dpar="mu", predict="prediction")`. Previously, when `predict="mu"` or any other dpar, `predict` was always set to `"expectation"`, and if you passed `dpar` directly, this did no longer work.

DominiqueMakowski:
yes that makes sense, I'll test it!

PR #1123, opened 2025-08-25
Update format_p.R to remove leading zeros consistently

ianhussey:
Previously, leading zeros were only removed when the value was truncated. `format_p(.02)` would return "p = 0.02" but `format_p(.00002)` would return "p < .001". APA style suggests always dropping leading zeros from p-values.  This commit adds a new argument, leading_zeros, which defaults to FALSE. When FALSE, a `gsub()` call removes the leading zero before the decimal_seperator.

### DIFF
--- a/R/format_p.R
+++ b/R/format_p.R
@@ -17,6 +17,9 @@
 #'   If `"scientific"`, control the number of digits by adding the value as
 #'   a suffix, e.g.m `digits = "scientific4"` to have scientific notation
 #'   with 4 decimal places.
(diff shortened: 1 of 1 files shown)


strengejacke:
Thanks! A couple of comments. Could you also increase the dev-version number in the `DESCRIPTION` file? And please add a small test.

strengejacke:
Don't worry, I can add tests. Could you please look at my comments?

bwiernik:
For consistency, I think we should include a leading zero in the truncated format `"p < 0.001"` when `leading_zero = TRUE`

PR #1159, opened 2026-01-16
Make is_empty_object faster

### LINKED ISSUE #1158: Make is_empty_object faster
Hello,

After digging a rabbit hole from slow table creation with modelsummary, I ended up here.

Long story short, I'm estimating 12 models using fixest and ~1M observations stored in data.tables. Then I'm using modelsummary() to create my tables. The whole function takes 42s to run, which is quite long to mostly format tables.

It appears that 20s are spent in is_empty_object() due to the as.data.frame() calls that copy the whole data multiple times across the whole run.

https://github.com/easystats/insight/blob/9771a107689fd9c7b7163a97cfdb52782228c99e/R/utils_object.R#L12

From my understanding, these calls to as.data.frame() are done to handle some edge cases, but do you think there would be a faster way to do it ?

Moreover, it appears that in the case of tibbles, the data is copied two times.

Here is a small reproductible example of my issue

```R
N <- 10^8
K <- 10^3

X <- rnorm(N)
Y <- 1 + 2 * X + rnorm(N)

dt <- data.table::data.table(X, Y)

model <- fixest::feols(Y ~ X, data = dt)

model |>
  modelsummary::modelsummary(gof_map = NA) |>
  profvis::profvis()

# Results on my computer
# modelsummary = 1470ms
# is_empty_object = 790ms
```

I commented out the as.data.frame() call to see the difference

```R
my_is_empty_object <- function (x) 
{
  flag_empty <- FALSE
  if (inherits(x, c("tbl_df", "tbl"))) 
    x <- as.data.frame(x)
  if (inherits(x, "data.frame")) {
    # x <- as.data.frame(x)
    flag_empty <- TRUE
      i <- 0
      while (flag_empty && i <= ncol(x)) {
        i <- i + 1
        flag_empty <- all(is.na(x[[i]]))
      }
    }
  }
  else if (is.list(x) && length(x) > 0) {
    x <- .safe(compact_list(x), x)
  }
  else if (!is.null(x)) {
    x <- stats::na.omit(x)
  }
  isTRUE(flag_empty) || length(x) == 0 || is.null(x) || isTRUE(nrow(x) == 
                                                                 0) || isTRUE(ncol(x) == 0)
}

environment(my_is_empty_object) <- asNamespace('insight')
assignInNamespace("is_empty_object", my_is_empty_object, ns = "insight")

model |>
  modelsummary::modelsummary(gof_map = NA) |>
  profvis::profvis()

# Results on my computer
# modelsummary = 720ms
# is_empty_object = 10ms
````

strengejacke:
Fixes #1158

### DIFF
--- a/tests/testthat/test-lmer.R
+++ b/tests/testthat/test-lmer.R
@@ -476,7 +476,7 @@ suppressMessages({
     Reaction ~
       (1 |
         mygrp / mysubgrp) +
-        (1 | Subject),
+      (1 | Subject),
     data = sleepstudy
   )
 
@@ -585,12 +585,12 @@ test_that("get_predicted_ci: warning when model matrix and varcovmat do not matc
   mod <- suppressMessages(lme4::lmer(
     weight ~
       1 +
-        Time +
-        I(Time^2) +
-        Diet +
-        Time:Diet +
-        I(Time^2):Diet +
-        (1 + Time + I(Time^2) | Chick),
+      Time +
+      I(Time^2) +
+      Diet +
+      Time:Diet +
+      I(Time^2):Diet +
+      (1 + Time + I(Time^2) | Chick),
     data = ChickWeight
   ))
   newdata <- ChickWeight[ChickWeight$Time %in% 0:10 & ChickWeight$Chick %in% c(1, 40), ]
--- a/tests/testthat/test-plm.R
+++ b/tests/testthat/test-plm.R
@@ -140,12 +140,12 @@ withr::with_package(
     mar <- suppressWarnings(plm::pgmm(
       log(emp) ~
         lag(log(emp), 1:2) +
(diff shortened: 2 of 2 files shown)


strengejacke:
And: Has anybody an idea why these snapshot tests fail?
https://github.com/easystats/insight/actions/runs/21078063446/job/60624846511?pr=1159

etiennebacher:
> I don't understand why these fail.

Only happens due to `testthat` 3.3.2 (updated a few days ago). I'll try to narrow it down.

strengejacke:
Thanks for figuring this out! I realized that panelr was removed from CRAN, but I think we had this issue also with other packages that still were/are on CRAN. Especially the change of requirements for the Matrix package were causing some troubles.

PR #1161, opened 2026-01-20
tidymodels methods: find_variables & get_data

vincentarelbundock:
This might already be fit for purpose, but I have not played with it a lot, so marking as "work in progress" (draft) for now.

### DIFF
--- a/tests/testthat/test-export_table.R
+++ b/tests/testthat/test-export_table.R
@@ -1,3 +1,5 @@
+skip_on_ci()
+
 d <- data.frame(a = c(1.3, 2, 543), b = c("ab", "cd", "abcde"), stringsAsFactors = FALSE)
 
 test_that("export_table", {


strengejacke:
Thanks! I have to submit an update by 31st January, I think. Do you think this could be merged before that deadline and be part of that submission, or would you like to have a longer test phase for this PR?

vincentarelbundock:
Yes, i think we can definitely make this work before the 31st.

What's the deal with the current failures? At first glance some of them look unrelated. Should I expect everything to pass?

R/find_predictors.R
```diff
@@ -194,6 +194,74 @@ find_predictors.default <- function(
...
+      return(NULL)
+    }
+
+    predictor_vars <- var_info$variable[var_info$role == "predictor"]
+    if (length(predictor_vars) == 0) {
+      return(NULL)
+    }
+
+    l <- list(conditional = predictor_vars)
+    return(if (flatten) unique(unlist(l, use.names = FALSE)) else compact_list(l))
```
strengejacke: Ok, this is crazy :rocket:

strengejacke (approved):

strengejacke (changes requested):
Can we remove the pipes from the tests?

PR #1164, opened 2026-02-01
Prepare CRAN release

strengejacke:


etiennebacher:
I don't see any related changes in R-source recently, and `tweedie` hasn't changed since 2022 so I have no idea. Did you try running win-devel? My guess is that it's a random failure and shouldn't threaten the submission.

strengejacke:
I'm "faking" the function call now (https://github.com/easystats/insight/pull/1164/changes/2b45004a0cc125d9e97cdf71d6a8c2990ee1ea98), let's see if win-builder complains.

Issue #1165, opened 2026-02-04
`find_parameters.gam` can be very slow for models with random effect smooths


gavinsimpson:
`find_parameters.gam()` calls `summary()` on the model to extract the `s.table` component. Unfortunately, for models with random effect smooths, the tests performed on model terms by `summary.gam()` are expensive computationally for "random effect" terms, increasingly so as $n$ gets large or the number and complexity of the random effects grows. You can turn off these tests with the argument `re.test = FALSE` but this has the unfortunate side effect of excluding any random effect terms from the summary, so they'd get missed by `find_parameters()`.

It is not necessary to call `summary()` to extract the labels of the smooths as mgcv knows them; they are stored in a logical way inside the `$smooth` component of the `gam()` fit.

I would suggest replacing the lines:

https://github.com/easystats/insight/blob/0d1d835f786c42c98d070af8f20c08f464f1541e/R/find_parameters_gam.R#L49

and

https://github.com/easystats/insight/blob/0d1d835f786c42c98d070af8f20c08f464f1541e/R/find_parameters_gam.R#L52

with a single call to replace L52:

```r
par$smooth_terms <- vapply(x[["smooth"]], FUN = `[[`, FUN.VALUE = character(1), "label")
```

If you are not comfortable digging in the internals of a gam object (these haven't changed in over a decade and are so baked into the code of mgcv that they are unlikely to ever change, but I am not the developer of mgcv, so take that as you will) you could use `mgcv::smooth.info` as an intermediary

```r
st <- lapply(x$smooth, mgcv::smooth.info)
par$smooth_terms <- vapply(st, FUN = `[[`, FUN.VALUE = character(1), "label")
```

This probably needs a little checking as `$smooth` can be and empty list if the model doesn't contain any smooths.

If you're interested in making this change, I can prepare a PR implementing it (with the necessary check for no smooths etc.)

PR #1166, opened 2026-02-05
Avoid calling `summary()` on a `gam()` model

gavinsimpson:
This PR addresses issue #1165 by avoiding calling `summary()` on a `gam()` model to identify the names of model parameters. `summary()` can be slow for GAMs with random effects.

This PR does two things:

* `R/find_parameters.R` is modified to directly extract from the model object the labels of any smooths in the model, and
* `tests/testthat/test-gam.R` is modified slightly through addition of a GAM with only parametric terms and a test of how `find_parameters()` works in that case.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Type: Package
 Package: insight
 Title: Easy Access to Model Information for Various Model Objects
-Version: 1.4.6
+Version: 1.4.6.1
 Authors@R:
     c(person(given = "Daniel",
              family = "Lüdecke",


gavinsimpson:
@strengejacke I can do that; I didn't think it was warranted in this case and the guidelines implied (to me at least) that this was for user-visiable changes and this doesn't change anything (which was the point; unless rare performance issues are "user-visible").

strengejacke:
True. But I think more or less "substantial" performance-issues can justify a version increase.

PR #1172, opened 2026-02-20
[Feature] get_simulated

### LINKED ISSUE #1171: Option to simulate predictions based on `simulate()`

DominiqueMakowski:
Fixes #1171


DominiqueMakowski:
My feeling is that get_simulated should be useable from within get_predicted (which would enable seemless integration at higher levels, modelbased etc). The question is how to do it in a conceptually consistent and logical manner.

@bwiernik: *"specifying bootstrap should be in ci_method not predict"*

Current behaviour:
- (default) `get_predicted.lm(m, iterations=NULL, ci_method=NULL)`: `predict.lm` + regular CI
-  `get_predicted.lm(m, iterations=100, ci_method=NULL)`: boostrapped `predict.lm` + CI from these draws
-  `get_predicted.lm(m, iterations=NULL, ci_method="bootstrapped")`: not existing

*(Note: for LMs, predict type doesn't matter here)*

New proposal:
- default doesn't change
-  `get_predicted.lm(m, predict="expectation", iterations=100, ci_method=NULL)`: bootstrapped link-predictions + regular CI (NOT using these draws for CI, just analytically compute so that the CI
-  `get_predicted.lm(m, predict="prediction", iterations=100, ci_method=NULL)`: get_simulated() + CI from these draws
-  `get_predicted.lm(m, predict="expectation", iterations=NULL, ci_method="bootstrapped")`:  `predict.lm` + CI computed from draws with iterations get set to a default number (e.g., 1000). I.e., CI is basde on draws but predictions are still deterministic
- `get_predicted.lm(m, predict="prediction", iterations=100, ci_method="bootstrapped")`: 🤷 one option is to decouple the predictions draws (based on simulate) from the CI computation (based on boostrap), but bootstrap what? expectations? predictions?


@mattansb @strengejacke

strengejacke:
I wouldn't add `"bootstrapped"` as `ci_method` option, what option / what kind of CIs would that be?

DominiqueMakowski:
I suppose what you say also makes sense. Formally, the hiatus is between these two positions:
- The CI refers to the credibility of the draws when multiple draws are present, regardless of what these draws are
- The CI refers to the CI of the estimation, which can be in some cases decoupled bfrom the predictions (eg bootstrapped CI + non-bootsrapped predictions)

strengejacke:
> The CI refers to the CI of the estimation, which can be in some cases decoupled bfrom the predictions (eg bootstrapped CI + non-bootsrapped predictions)

Are there situations where you really want these? They're probably no good "dispersion" or "uncertainty" estimates and - in extreme cases - may not even include the point estimate (predicted value)?

strengejacke:
Why do we have all these

```r
#' @export
get_simulated.polr <- get_simulated.default
```

?

strengejacke:
We should check whether we have binary, ordinal or categorical outcomes. In this case, `simulate()` doesn't work well.

``` r
library(insight)
model <- glm(vs ~ am + wt, data = mtcars, family = "binomial")

# fails, because we call `simulate()` on the entire model,
# but we do not filter data
out <- get_simulated(
  model,
  iterations = 2,
  seed = 123,
  data = insight::get_datagrid(model, "am")
)
#> Error in `dim(val) <- c(n, iterations)`:
#> ! dims [product 4] do not match the length of object [64]

# works, but anyway - simulated values are only 0 or 1, we can't calculate
# a probability or useful intervals
out <- get_simulated(
  model,
  iterations = 2,
  seed = 123
)
out
#>                     iter_1 iter_2
#> Mazda RX4                0      0
#> Mazda RX4 Wag            0      0
#> Datsun 710               1      1
#> Hornet 4 Drive           0      1
#> Hornet Sportabout        1      1
#> Valiant                  0      0
#> Duster 360               0      0
#> Merc 240D                0      1
#> Merc 230                 1      1
#> Merc 280                 0      0
#> Merc 280C                1      0
#> Merc 450SE               0      0
#> Merc 450SL               0      0
#> Merc 450SLC              0      0
#> Cadillac Fleetwood       0      0
#> Lincoln Continental      0      0
#> Chrysler Imperial        0      0
#> Fiat 128                 1      0
#> Honda Civic              1      1
#> Toyota Corolla           1      1
#> Toyota Corona            1      1
#> Dodge Challenger         1      0
#> AMC Javelin              0      0
#> Camaro Z28               1      0
#> Pontiac Firebird         0      0
#> Fiat X1-9                1      1
#> Porsche 914-2            1      0
#> Lotus Europa             1      1
#> Ford Pantera L           0      0
#> Ferrari Dino             0      0
#> Maserati Bora            0      0
#> Volvo 142E               1      0
```

<sup>Created on 2026-02-21 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

strengejacke:
Ok, should work for binomial now. But we have to add internal simulate-function for other glm-families, too (to make it work with data grids)

DominiqueMakowski:
> works, but anyway - simulated values are only 0 or 1, we can't calculate a probability or useful intervals

I think for logistic models that'd be the expected behaviour, same as posterior_predict also returns zeros and ones

DominiqueMakowski:
Sorry I disapeared lol 

> Do we also want to include the data? 

🤷 up to you

I still expect to use this function somehow through get_predicted rather than on its own anyway

good stuff for the rest 👍

PR #1208, opened 2026-07-21
Add "weighted" argument for datagrids

### LINKED ISSUE #1207: Add "weighted" argument for datagrids
https://github.com/easystats/modelbased/issues/642

The goal is to enable faster marginal processing: instead of duplicating data(grids), we directly provide the weighting associated with each combination.

- Arguments adds a "Weights" column (see also [here](https://marginaleffects.com/bonus/performance.html#solution-c-weighted-grid))
- For `numerics`, add the option to bin them (how many bins controlled by `length`?)
- Treat binned numerics as factors in the weighting process

This Weight column can be then used in modelbased/marginaleffects. 

But how would that work with random factors? We need some proof of concept toy examples to assess whether it is a useful feature to work on or not

strengejacke:
Fixes #1207

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Type: Package
 Package: insight
 Title: Easy Access to Model Information for Various Model Objects
-Version: 1.5.2.4
+Version: 1.5.2.5
 Authors@R:
     c(person(given = "Daniel",
              family = "Lüdecke",
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,9 +2,9 @@
 
 ## Changes
 
-* `get_datagrid()` gains a `weight` argument, to create a smaller representation
+* `get_datagrid()` gains a `weighted` argument, to create a smaller representation
   of large data grids, where multiple unique combinations of predictors are
-  included only once, and a new `weight` variable indicates how often each
+  included only once, and a new `weight` column indicates how often each
   combination appears in the original data.
 
 * `format_ci()` gains a `separator` argument, to customize the separator for
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,11 @@
 
 ## Changes
 
+* `get_datagrid()` gains a `weight` argument, to create a smaller representation
+  of large data grids, where multiple unique combinations of predictors are
+  included only once, and a new `weight` variable indicates how often each
+  combination appears in the original data.
+
 * `format_ci()` gains a `separator` argument, to customize the separator for
   lower and upper CI values.
 
--- a/R/get_datagrid.R
+++ b/R/get_datagrid.R
@@ -707,6 +707,7 @@ get_datagrid.default <- function(
   by = "all",
   factors = "reference",
   numerics = "mean",
+  weighted = FALSE,
   preserve_range = TRUE,
   reference = x,
   include_smooth = TRUE,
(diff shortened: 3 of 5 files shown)


strengejacke:
@DominiqueMakowski 

- Which name should the argument have? Currently, it's `weight`
- Should it be a logical? If `TRUE`, create weighted grid based on variables in `by`?
- For now, we could set numerics to their mean, and later indeed bin them, as you suggested?
- If we have real "weights" in the data, e.g. sampling weights or similar, we should use the sum of the weights and multiply them with the created `weight` variable, to take sampling/design weights into account when creating a weighted data grid - but how can users tell `get_datagrid()` that there is an additional weight-variable?

DominiqueMakowski:
- `weighted` ?
- Yes I think so
- Yes
- Interesting idea, I am not fully sure whether there's any caveat there but yes, perhaps we could allo weighted to take a column name as an input to then adjust weights (although it might be overkill for now? maybe best to start simple)

strengejacke:
I think including weights is not that difficult. We just need to multiply `nrow(s)` with something like `sum(s$weights_var)` in this line:
https://github.com/easystats/insight/blob/072906cca6c02152f163dbbc2895301d62750380/R/get_datagrid.R#L1311

strengejacke:
The idea of this PR is basically to get a "reduced" full data set, right? I think therefore, when `weighted` is provided, we have to ignore other arguments like `length` or `range` for now. Or do you have an idea how to include / use these arguments?

DominiqueMakowski:
``` r
insight::get_datagrid(iris, by = c("Sepal.Length", "Species"), length = 3)
#>   Sepal.Length    Species Sepal.Width Petal.Length Petal.Width
#> 1          4.3     setosa    3.057333        3.758    1.199333
#> 2          6.1     setosa    3.057333        3.758    1.199333
#> 3          7.9     setosa    3.057333        3.758    1.199333
#> 4          4.3 versicolor    3.057333        3.758    1.199333
#> 5          6.1 versicolor    3.057333        3.758    1.199333
#> 6          7.9 versicolor    3.057333        3.758    1.199333
#> 7          4.3  virginica    3.057333        3.758    1.199333
#> 8          6.1  virginica    3.057333        3.758    1.199333
#> 9          7.9  virginica    3.057333        3.758    1.199333
```

<sup>Created on 2026-07-23 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

Well, as far as I can see, this "type" of datagrids is *not* used in marginal estimations (i.e., marginal means do not rely on predictions at arbitrary levels do they? - unless we include one in `by`). In which case I am not sure what's the best approach, whether to ignore it or not. 

@vincentarelbundock We are thinking of implementing a "reduced" version of datagrids, with the endgoal of making our modelbased's marginal means faster by passing a weighed grid to marginaleffects (following your example). 

Do you have any intuition on how do work with weights when making predictions on arbitrary/new values of continuous variables?

vincentarelbundock:
I don't understand the statistical motivation. Normally, weights are designed to target a specific population. It doesn't make much sense to me think of them as context-free for "arbitrary grids".

Under the hood in marginaleffects, we sometimes use weights to speed up things when we target the empirical grid.

Also, please try the GitHub version. Many operations are now 10x faster.

vincentarelbundock:
Just to clarify: `marginaleffects` automatically reduces large grids and uses weights when it's possible and makes statistical sense (categorical-only predictors).

You probably won't get much speedup when using the `marginaleffects` backend. (There might be minor benefit from not calling `unique()` on the grid.)

DominiqueMakowski:
@strengejacke then perhaps we should not bother for now until there is a more established reason to

@vincentarelbundock The (vague) idea was still in the context of cases like this: https://github.com/easystats/modelbased/issues/639#issue-4786735572 ie large input dataframes, to run the marginal estimation on compact representations, but if you say that it's already internally optimized at marginaleffects level then I guess there's no speedups to expect from that.

vincentarelbundock:
Right. I thought about this a fair bit back in the day, and it really only works *perfectly* when a model only includes categorical predictors. And then, it only speeds up to the extent that the number of unique combinations of levels is smaller than your original grid.

For a single call, you won't get any speedup by doing this ahead of time with `modelbased`. If you do the same call repeatedly, then there might be gains in pre-computing the smaller grid and the appropriate set of weights. But, personally, I would not mix these two complementary concerns. There is already a `wts` and a `newdata` argument. Combining them into a new "weighted grid" concept seems like adding a confusing layer of abstraction on top of an already clear API. But YMMV, of course.

And yeah, try the github version when you have a chance. I think it'll be meaningfully faster and use less memory.

strengejacke:
``` r
library(insight)
data(penguins)

# one factor
get_datagrid(penguins, "species", weighted = TRUE)
#>     species weight
#> 1    Adelie    152
#> 2 Chinstrap     68
#> 3    Gentoo    124

# one factor, filtered
get_datagrid(penguins, "island=c('Biscoe', 'Dream')", weighted = TRUE)
#>   island weight
#> 1 Biscoe    168
#> 2  Dream    124

# two factors
get_datagrid(penguins, c("species", "island"), weighted = TRUE)
#>     species    island weight
#> 1    Adelie    Biscoe     44
#> 2    Gentoo    Biscoe    124
#> 3    Adelie     Dream     56
#> 4 Chinstrap     Dream     68
#> 5    Adelie Torgersen     52

# two factors, filtered
get_datagrid(
  penguins,
  c("species=c('Adelie', 'Gentoo')", "island=c('Biscoe', 'Dream')"),
  weighted = TRUE
)
#>   species island weight
#> 1  Adelie Biscoe     44
#> 2  Gentoo Biscoe    124
#> 3  Adelie  Dream     56

# two factors, one numeric
get_datagrid(penguins, c("species", "island", "body_mass"), weighted = TRUE)
#>     species    island weight body_mass
#> 1    Adelie    Biscoe     44  4201.754
#> 2    Gentoo    Biscoe    124  4201.754
#> 3    Adelie     Dream     56  4201.754
#> 4 Chinstrap     Dream     68  4201.754
#> 5    Adelie Torgersen     52  4201.754

# auto-detection
get_datagrid(penguins, weighted = TRUE)
#>      species    island    sex weight bill_len bill_dep flipper_len body_mass
#> 1     Adelie    Biscoe female     22   43.922   17.151     200.915  4201.754
#> 2     Gentoo    Biscoe female     58   43.922   17.151     200.915  4201.754
#> 3     Adelie     Dream female     27   43.922   17.151     200.915  4201.754
#> 4  Chinstrap     Dream female     34   43.922   17.151     200.915  4201.754
#> 5     Adelie Torgersen female     24   43.922   17.151     200.915  4201.754
#> 6     Adelie    Biscoe   male     22   43.922   17.151     200.915  4201.754
#> 7     Gentoo    Biscoe   male     61   43.922   17.151     200.915  4201.754
#> 8     Adelie     Dream   male     28   43.922   17.151     200.915  4201.754
#> 9  Chinstrap     Dream   male     34   43.922   17.151     200.915  4201.754
#> 10    Adelie Torgersen   male     23   43.922   17.151     200.915  4201.754
#>        year
#> 1  2008.029
#> 2  2008.029
#> 3  2008.029
#> 4  2008.029
#> 5  2008.029
#> 6  2008.029
#> 7  2008.029
#> 8  2008.029
#> 9  2008.029
#> 10 2008.029

# errors
get_datagrid(penguins, "test", weighted = TRUE)
#> Error:
#> ! Variable `test` was not found in the data. Please check the spelling.

get_datagrid(penguins, "idland", weighted = TRUE)
#> Error:
#> ! Variable `idland` was not found in the data. Did you mean "island"?

get_datagrid(penguins, "body_mass", weighted = TRUE)
#> Error:
#> ! No factors were specified in `by`, which is required to create a
#>   weighted data grid.

get_datagrid(penguins, "island", weighted = "weights")
#> Error:
#> ! The variable `weights` does not exist in the data.

d <- penguins
set.seed(123)
d$weights <- abs(rnorm(nrow(d), 1, 0.2))

get_datagrid(d, "island", weighted = "weight")
#> Error:
#> ! The variable `weight` does not exist in the data. Did you mean
#>   "weights"?

# double weighting
get_datagrid(d, "island", weighted = "weights")
#>      island    weight
#> 1    Biscoe 167.05601
#> 2     Dream 127.47271
#> 3 Torgersen  52.24464
```

<sup>Created on 2026-07-23 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

PR #1209, opened 2026-07-23
Fix `null_model()` for inline offsets with nested parens and trailing terms

raffaelemancuso:
## Problem

`null_model()` errors for a model whose offset is written inline in the formula
as a **nested-parenthesis expression followed by further terms**:

```r
library(glmmTMB)
m <- glmmTMB(
  y ~ x + offset(log(exposure)) + factor(year),
  data = d, family = nbinom2()
)
insight::null_model(m)
#> Error in str2lang(offset_term) :
#>   <text>:1:19: unexpected ')'
#>   1: log(exposure))
```

This also breaks every downstream consumer that builds a null model, e.g.
`performance::r2_mcfadden()` / `r2_nagelkerke()` / `r2_coxsnell()` and therefore
`performance::model_performance()`.

## Cause

`.grep_offset_term()` extracted the offset from the *deparsed* formula with a
paren-blind regex:

```r
gsub("(.*)offset\\((.*)\\)(.*)", "\\2", f)
```

That assumes the closing paren of `offset(` is the last `)` in the string. When
the offset expression itself contains parentheses (`offset(log(exposure))`) and
other terms follow it, the offset's `)` is no longer the last one, so the regex
captures an unbalanced, unparseable substring
(`log(exposure)) + factor(year`), and `str2lang()` fails.

## Fix

Extract the offset by walking the formula's language tree for the `offset()`
call and deparsing its argument. This is correct regardless of nested
parentheses or term order, and leaves the previously-working cases unchanged.

## Tests

Added a regression test in `test-null_model.R` (inline nested-paren offset with
a trailing `factor()` term) asserting `null_model()` builds without error, keeps
the offset, and matches the equivalent model fit with the `offset =` argument.
`test-null_model.R` and `test-offset.R` pass.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,12 @@
 # insight (devel)
 
+## Bug fixes
+
+* `null_model()` no longer fails for models whose offset is written inline in
+  the formula with a nested-parenthesis expression followed by further terms
+  (e.g. `y ~ x + offset(log(exposure)) + factor(year)`). The offset term is now
+  extracted from the formula's language tree instead of by a paren-blind regex.
+
 ## Changes
 
 * `format_ci()` gains a `separator` argument, to customize the separator for
--- a/R/null_model.R
+++ b/R/null_model.R
@@ -315,21 +315,43 @@ null_model.glmmadmb <- null_model.glmmTMB
       # its argument. A regex on the deparsed formula cannot reliably extract
       # the offset expression when that expression contains nested parentheses
       # (e.g. `offset(log(x))`) or when further terms follow the offset in the
-      # formula (e.g. `... + offset(log(x)) + factor(year)`): the closing paren
-      # of `offset(` is then no longer the last `)` in the string, so the old
-      # regex captured an unbalanced, unparseable substring.
-      offset_expr <- NULL
-      find_offset_call <- function(e) {
-        if (!is.null(offset_expr) || !is.call(e)) {
+      # formula (e.g. `... + offset(log(x)) + factor(year)`): the closing paren of
+      # `offset(` is then no longer the last `)` in the string, so the old regex
+      # captured an unbalanced, unparseable substring.
+      .find_offset_call <- function(e) {
+        # If the current element is not a call (e.g., a symbol or a constant),
+        # it cannot be an offset() call, so we return NULL.
+        if (!is.call(e)) {
           return(NULL)
         }
+
+        # Check if the current call is `offset(...)`
+        # We ensure it's exactly the function 'offset' and has at least one argument.
         if (identical(e[[1L]], as.name("offset")) && length(e) >= 2L) {
-          offset_expr <<- e[[2L]]
-          return(NULL)
+          # Return the argument passed to offset()
+          return(e[[2L]])
         }
-        for (i in seq_along(e)) find_offset_call(e[[i]])
+
+        # Recursively walking the syntax tree. If it's not the offset call,
+        # recursively check all components of the current call. We iterate over
+        # all elements (e.g., in `y ~ x + offset(z)`, `e` is a call to `~` with
+        # arguments `y` and `x + offset(z)`)
+        for (i in seq_along(e)) {
+          res <- .find_offset_call(e[[i]])
+          # If a non-NULL result is found in any of the child nodes, immediately
+          # return it up the call stack.
+          if (!is.null(res)) {
+            return(res)
+          }
+        }
+
+        # If the offset call is not found in this branch, return NULL
+        return(NULL)
       }
-      find_offset_call(f)
+      # Executing the search and formatting the result. Start the recursive
+      # search on the formula 'f'
+      offset_expr <- .find_offset_call(f)
+
       if (is.null(offset_expr)) {
         NULL
       } else {
--- a/R/null_model.R
+++ b/R/null_model.R
@@ -307,13 +307,34 @@ null_model.glmmadmb <- null_model.glmmTMB
 .grep_offset_term <- function(model_formula) {
   tryCatch(
     {
-      f <- safe_deparse(model_formula$conditional)
-      if (grepl("offset(", f, fixed = TRUE)) {
-        out <- gsub("(.*)offset\\((.*)\\)(.*)", "\\2", f)
-      } else {
-        out <- NULL
+      f <- model_formula$conditional
+      if (is.null(f)) {
+        return(NULL)
       }
-      out
+      # Walk the formula's language tree to find the offset() call and return
+      # its argument. A regex on the deparsed formula cannot reliably extract
+      # the offset expression when that expression contains nested parentheses
+      # (e.g. `offset(log(x))`) or when further terms follow the offset in the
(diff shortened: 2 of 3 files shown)


strengejacke:
Thanks a lot! I would like to avoid `<<-`, and maybe directly let `find_offset_call()` return the offset-expression. I have revised the code and would make following suggestion, which still passes all tests. Let me know what you think:

```r
      f <- model_formula$conditional
      if (is.null(f)) {
        return(NULL)
      }
      # Walk the formula's language tree to find the offset() call and return
      # its argument. A regex on the deparsed formula cannot reliably extract
      # the offset expression when that expression contains nested parentheses
      # (e.g. `offset(log(x))`) or when further terms follow the offset in the
      # formula (e.g. `... + offset(log(x)) + factor(year)`): the closing paren of
      # `offset(` is then no longer the last ``)` in the string, so the old regex
      # captured an unbalanced, unparseable substring.
      .find_offset_call <- function(e) {
        # If the current element is not a call (e.g., a symbol or a constant),
        # it cannot be an offset() call, so we return NULL.
        if (!is.call(e)) {
          return(NULL)
        }

        # Check if the current call is `offset(...)`
        # We ensure it's exactly the function 'offset' and has at least one argument.
        if (identical(e[[1L]], as.name("offset")) && length(e) >= 2L) {
          # Return the argument passed to offset()
          return(e[[2L]])
        }

        # Recursively walking the syntax tree If it's not the offset call,
        # recursively check all components of the current call. We iterate over
        # all elements (e.g., in `y ~ x + offset(z)`, `e` is a call to `~` with
        # arguments `y` and `x + offset(z)`)
        for (i in seq_along(e)) {
          res <- .find_offset_call(e[[i]])
          # If a non-NULL result is found in any of the child nodes, immediately
          # return it up the call stack.
          if (!is.null(res)) {
            return(res)
          }
        }

        # If the offset call is not found in this branch, return NULL
        return(NULL)
      }
      # Executing the search and formatting the resultStart the recursive search
      # on the formula 'f'
      offset_expr <- .find_offset_call(f)

      if (is.null(offset_expr)) {
        NULL
      } else {
        safe_deparse(offset_expr)
      }
```

raffaelemancuso:
@strengejacke Hi, if you commit to this or another PR I can test it.

strengejacke:
Sure, just committed to this PR. It's basically an adaption of your code, mainly to avoid `<<-` because I think this might be an issue when submitting to CRAN.

raffaelemancuso:
Sorry for the mess.

Your modifications looks good, I added proper testthat for formulas in which the offset is followed by another term with round brackets in it.

Formulas which explicit the `stats` package for the offset, like:

```
y ~ stats::offset(log(a)) + factor(yr)
```

Are worth checking because they return a diffrent-than-1 coefficient even in base R

strengejacke:
> Are worth checking because they return a diffrent-than-1 coefficient even in base R

That's strange. What does this mean for this PR? Can it be merged, or is there something you would like to add?

raffaelemancuso:
> > Are worth checking because they return a diffrent-than-1 coefficient even in base R
> 
> That's strange. What does this mean for this PR? Can it be merged, or is there something you would like to add?

For this PR it means that the PR replicates R behavior and thus can be merged.

But maybe there is a bug in R? Or is it supposed to work like that? Can you check that behavior?

strengejacke:
Ok will check later (currently travelling and only occasionally at my laptop)

raffaelemancuso:
Well, that turned out to be how glmmTMB processes the formula. You are supposed to use `offset`, never `stats::offset`. So I think we are good.

strengejacke:
@bbolker maybe of interest for you?

See https://github.com/easystats/insight/pull/1209#issuecomment-5089528467

bbolker:
Do we need all of this machinery?  Using the machinery from [reformulas](https://github.com/bbolker/reformulas) (if you're willing to import it):

```r
library(reformulas)
null_model <- function(formula) {
    if (length(formula) < 3) {
        stop("formula must be two-sided (have a response)")
    }
    response <- formula[[2]]
    bars <- findbars(formula)
    if (is.null(bars)) {
        return(as.formula(bquote(.(response) ~ 1),
                           env = environment(formula)))
    }
    rhs <- sumTerms(lapply(bars, makeOp, quote(`(`)))
    as.formula(bquote(.(response) ~ 1 + .(rhs)),
               env = environment(formula))
}

f <- y ~ x + offset(log(exposure)) + factor(year)
null_model(f)
f2 <- y ~ x + offset(log(exposure)) + factor(year) + ar1(junk|other) + (1 + x|f/g)
## it does expand the `/` term (not sure if that's a problem?
y ~ 1 + (junk | other) + (1 + x | g:f) + (1 + x | f)
```

(is there a problem with using `stats::offset` in a glmmTMB formula? If so, please report it as a glmmTMB issue **update**: a little experimentation shows that all the low-level machinery in R processing (`model.frame()`, `terms()` etc.) recognizes only `offset`, not `stats::offset`. Don't know if there's a documentation fix for that, although it's not a problem anyone has ever reported before now ...)

FWIW I'm a little bit surprised that there are so many individual S3 methods for `null_model`. Wouldn't most model types be OK with using the `formula()` S3 method to extract the formula and then processing it in a generic way, i.e. could you save a lot of complexity by making that the default method? (And maybe adding a `formula()` S3 method for model types that were missing it ...)

strengejacke:
> I'm a little bit surprised that there are so many individual S3 methods for null_model

I think it's mostly due to different `update()` methods, where the different model classes need different extras arguments, plus the glmmTMB method for the extra zero inflation handling.

raffaelemancuso:
> (is there a problem with using stats::offset in a glmmTMB formula? If so, please report it as a glmmTMB issue update: a little experimentation shows that all the low-level machinery in R processing (model.frame(), terms() etc.) recognizes only offset, not stats::offset. Don't know if there's a documentation fix for that, although it's not a problem anyone has ever reported before now ...)

It looks like you are supposed to always use `offset` not `stats::offset`. See [here](https://github.com/therneau/survival/issues/199#issuecomment-1148162483) for instance.

This is documented in the [help page for `formula`](https://stat.ethz.ch/R-manual/R-devel/library/stats/html/formula.html):

> Most model-fitting functions accept formulae with right-hand side containing calls to function [offset](https://stat.ethz.ch/R-manual/R-devel/library/stats/help/offset.html). These calls indicate terms with coefficient fixed to one. Some model-fitting functions recognize other “specials” like strata and cluster (typically by setting the specials argument of [terms.formula](https://stat.ethz.ch/R-manual/R-devel/library/stats/help/terms.formula.html)). Because specials are nothing more than syntax, disambiguation using [::](https://stat.ethz.ch/R-manual/R-devel/library/base/html/ns-dblcolon.html) or similar is nonsense (even if packages happen to define so-named functions) and the result is often a model that the user does not intend. For example, the formula y ~ stats::offset(w) + x does not specify a model in which the coefficient of w is fixed to one.

But I can't find anything else in the docs other than that

strengejacke:
Ok, thanks a lot, will merge this PR!

PR #1221, opened 2026-08-29
glmmTMB ordinal support

### LINKED ISSUE #1220: glmmTMB ordinal support
### Discussed in https://github.com/easystats/easystats/discussions/480

<div type='discussions-op-text'>

<sup>Originally posted by **jmgirard** August 26, 2026</sup>
Once we get https://github.com/glmmTMB/glmmTMB/issues/1323 resolved, I think we're ready to add support for ordinal models from glmmTMB: https://github.com/glmmTMB/glmmTMB/issues/514</div>

## Remaining issues to clarify / resolve

- [ ] `find_parameters()` and `get_parameters()` should (not) include the intercept?
- [ ] There are no intercept-thresholds, unlike in `ordinal::clm()` [this comment](https://github.com/easystats/insight/issues/1220#issuecomment-5461703164)?
- [ ] What is the scale of predictions? Response scale does not return probabilities yet [this comment](https://github.com/easystats/insight/issues/1220#issuecomment-5461712812)?

strengejacke:
Fixes #1220

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Type: Package
 Package: insight
 Title: Easy Access to Model Information for Various Model Objects
-Version: 1.5.3.1
+Version: 1.5.3.2
 Authors@R:
     c(person(given = "Daniel",
              family = "Lüdecke",
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,8 +1,17 @@
 # insight (devel)
 
-## Changes
+## New supported models
 
-* *insight* now supports the new `ordinal` family from *glmmTMB* models.
+* Support for models of class `glmmTMB` fitted with the new `ordinal()`
+  family (*glmmTMB* >= 1.1.15). As for `ordinal::clm()`, the thresholds are
+  returned as conditional parameters ahead of the slopes, with standard errors
+  and statistics obtained via the delta method (the thresholds are estimated on
+  an internal, softmax-parameterized scale), while the intercept, which is
+  fixed to zero by *glmmTMB*, is dropped. `get_predicted()` returns
+  per-category probabilities for `predict = "expectation"` and the most likely
+  category for `predict = "classification"`, and `get_variance()` uses the
+  distribution-specific residual variance of the latent scale, as for
+  `ordinal::clmm()`.
 
 ## Bug fixes
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,9 @@
 # insight (devel)
 
+## Changes
+
+* *insight* now supports the new `ordinal` family from *glmmTMB* models.
+
 ## Bug fixes
 
 * Fixed a bug where `get_statistic()` incorrectly extracted degrees of freedom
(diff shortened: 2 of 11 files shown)


jmgirard:
Thanks for starting this! Two things I noticed testing it on the `wine` data:

1. In `get_parameters()`, `out[-1, ]` runs after the thresholds are prepended, so it drops `1|2` and keeps `(Intercept)`.
2. The threshold rows of `vcov(full = TRUE)` are on glmmTMB's internal softmax scale, so the SEs don't match clm (0.99 vs 0.55 for `1|2`); they need the delta method (see glmmTMB/glmmTMB#1323).

I have a branch that handles both, plus `get_statistic()` (which currently errors on the row-count mismatch with `summary()`), `get_predicted()` via `type = "probs"` with clm-style intervals, `get_variance()` (latent residual variance, as for `clmm`), and tests against `clm`/`clmm`:

https://github.com/easystats/insight/compare/main...jmgirard:insight:glmmtmb-ordinal

Happy to open it as a PR, or you can cherry-pick from it into this one, whichever you prefer. One thing on the *parameters* side: `.se_fixed_effects_glmmTMB()` takes SEs from `coef(summary())`, which lacks the thresholds, so `model_parameters()` errors; falling back to `insight::get_varcov()` for the ordinal family would fix it.

strengejacke:
Great, thanks a lot! Let me check how `get_parameters.glmmTMB()` handles ordinal families, and if this is in line with `ordinal::clm()` and similar (i.e. thresholds are also returned).

Also, I noticed that `vcov()` for glmmTMB-ordinal returns fixed effects first, and then threshold estimates. We have to make sure this is in line with `get_parameters()`, so standard errors will correctly match in _parameters_.

jmgirard:
Good questions, luckily both should be covered in the branch already.

`get_parameters()` returns the thresholds first and then the slopes, with the same names and order as `ordinal::clm()`/`clmm()`; `find_parameters(m_tmb)` is `identical()` to `find_parameters(m_clm)` in `test-glmmTMB-ordinal.R`. For the covariance matrix, you're right that glmmTMB's `vcov()` puts the fixed effects first, and the threshold rows there are also on the internal softmax scale, so `get_varcov()` doesn't use it directly. It builds the delta-method matrix itself (`R/utils_glmmtmb_ordinal.R`) in the same order as `get_parameters()`, and the tests check the SEs against `clm()`/`clmm()`. On the *parameters* side, easystats/parameters#1253 takes SEs from `get_varcov()` matched by name, so `model_parameters()` matches `clmm()` column for column there as well.

On CI: the ordinal work is lint- and spelling-clean now; what's still red comes from main (the brms tests on Windows, styling on `test-brms_dpars.R`, and the older lints and spelling words). Happy to help clean those up in a separate PR if useful.

strengejacke:
Short note: Please don't focus too much on CI-tests, they're not fully reliable, AFAIK, and also yield some false-positives. Main formatting is now done by air, so not all lintr-comments must be followed (although there are always some useful comments).

strengejacke:
Let me start a copilot review, I'll also review this PR later in office.

strengejacke:
Great, thanks a lot! Looks good to me!

strengejacke:
Will run a final local check, and then merge

PR #1223, opened 2026-09-03
minor compatibility edit

DominiqueMakowski:
[cogmod](https://github.com/DominiqueMakowski/cogmod) is getting its first official release + CRAN upload, and so the family names are now prefixed with the package name to avoid issues

### DIFF
--- a/R/utils_model_info.R
+++ b/R/utils_model_info.R
@@ -554,5 +554,5 @@
 .is_rtchoice_family <- function(family_name) {
   !is.null(family_name) &&
     nzchar(family_name) &&
-    family_name %in% c("ddm", "lba", "lnr", "rdm")
+    family_name %in% c("cogmod_ddm", "cogmod_lba2", "cogmod_lnr", "cogmod_rdm")
 }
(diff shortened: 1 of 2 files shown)


strengejacke:
We would need to update the model in *circus* used [in this test](https://github.com/easystats/insight/blob/067ec0fe3dfe07c136b1aeab917c3457dcbd5491/tests/testthat/test-get_predicted.R#L892), or indeed keep the old names (at least for now) and add the new names, as copilot suggested.

DominiqueMakowski:
Let me update the circus models then I think

DominiqueMakowski:
wait where does the `m_ddm_1` comes from i can't find it

strengejacke:
Good question, I guess I just uploaded it after playing a bit with your package. 😬 

Let's comment out that test and add a new one later. I _think_ it was one of these models: https://dominiquemakowski.github.io/cogmod/articles/rt_models.html#recinormal-later

strengejacke:
I uploaded a new model to circus, documented it and updated the test here. Should all be ok now.

PR #1225, opened 2026-09-07
Don't match auxiliary parameter names fuzzily

### LINKED ISSUE #1224: find_auxiliary() hallucinates sigma parameters
``` r
m <- readRDS(url("https://raw.github.com/DominiqueMakowski/cogmod/main/vignettes/models/m_ddm.rds"))
insight::find_auxiliary(m)
#> Loading required namespace: rstan
#> [1] "boundary" "bias"     "ndt"      "sigma"

insight::find_parameters(m)
#> $conditional
#> [1] "b_Intercept"      "b_ConditionSpeed"
#> 
#> $bias
#> [1] "b_bias_Intercept"
#> 
#> $ndt
#> [1] "b_ndt_Intercept"
#> 
#> $boundary
#> [1] "b_boundary_Intercept"      "b_boundary_ConditionSpeed"
```

<sup>Created on 2026-09-07 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

The model has no sigma parameter, yet it is returned by `find_auxiliary()`

Somewhat the opposite of https://github.com/easystats/parameters/issues/1247, could be related

DominiqueMakowski:
Fixes #1224, and the related component grouping in easystats/parameters#1247.

## `find_auxiliary()` hallucinated `sigma`

The residual SD was detected by looking for the string `"sigma"` *anywhere* in the stan-model's parameter names:

```r
if (any(startsWith(fe, "sigma_") | grepl("sigma", fe, fixed = TRUE))) {
  out <- c(out, "sigma")
}
```

Custom brms-families whose auxiliary parameters merely *contain* that string therefore faked a `sigma` component. The `ddm()` family from *cogmod* estimates `sigmadrift`, `sigmabias` and `sigmandt` as constants, so:

```r
m <- readRDS(url("https://raw.github.com/DominiqueMakowski/cogmod/main/vignettes/models/m_ddm.rds"))
insight::find_auxiliary(m)
#> before: "boundary" "bias" "ndt" "sigma"
#> after:  "boundary" "bias" "ndt"
```

`sigma` is now matched exactly, and only when it isn't already covered by the model's `pforms`:

- `sigma` — univariate models with a constant residual SD
- `sigma_<response>` — multivariate models, validated against the actual response names (both the raw and brms' "cleaned" ones, e.g. `sigma_SepalLength` for `Sepal.Length`)
- `sigma1`, `sigma2`, … — mixture models

## `clean_parameters()` lumped `sigmabias` into `sigma`

Same root cause, different code path. `.get_stan_params()` mapped `find_parameters()` element names to components with a substring test, so the element *named* `sigmabias` matched `grepl("sigma", ...)` and its parameters were labelled `Component = "sigma"` — even though `find_parameters()` had correctly returned `sigma` and `sigmabias` as separate elements. That is what surfaces as easystats/parameters#1247; `parameters` itself needed no change, its output is downstream of `clean_parameters()`.

Component names are now compared for exact matches, after stripping the `_random` suffix that marks the group-level part of a component. The `grepl` branches for `conditional`, `priors`, `smooth_terms` and `dispersion` become redundant once that suffix is stripped, since they fall through to the identity default — so they are gone too, which removes the same fuzziness for those.

```r
m_lba <- readRDS(url("https://raw.github.com/DominiqueMakowski/cogmod/main/vignettes/models/m_lba1.rds"))
parameters::parameters(m_lba)
#> before: b_sigmabias_Intercept -> Component = sigma
#> after:  b_sigmabias_Intercept -> Component = sigmabias
```

Note this model genuinely has an exact `sigma` parameter (fixed to 1), so the `find_auxiliary()` fix alone did not address it; it still shows up as its own `sigma` component, correctly.

## Verification

- Old vs. new `find_auxiliary()` output compared across 19 downloadable brms models (univariate, mv with dotted response names, mixture, zi, zoib, ordinal, distributional `sigma ~ x`, `sigma` with random effects, custom `chocomini` family) — **identical in every case**.
- New `tests/testthat/test-brms_dpars.R` covers both bugs with lightweight `brmsfit` mockups (formula + stan parameter names), so no fitting or downloading is needed; the mockups reproduce the real models' output exactly. It also unit-tests the `.get_stan_params()` component mapping over the element names produced by all five classes that share the helper — `brmsfit`, `stanreg`, `stanfit`, `stanmvreg`, `bamlss` — since *rstanarm* isn't available locally to exercise those paths end-to-end.
- Full local test suite: no new failures. The three that remain (two lavaan `GFI` values in `test-export_table.R`, and `rt`/`RT` in `test-get_predicted.R:904`) reproduce unchanged on `main`.
- `parameters`' own `brms|stan|bayes` tests pass against this branch.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Type: Package
 Package: insight
 Title: Easy Access to Model Information for Various Model Objects
-Version: 1.5.4
+Version: 1.5.4.1
 Authors@R:
     c(person(given = "Daniel",
              family = "Lüdecke",
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,16 @@
+# insight (devel)
+
+## Bug fixes
+
+* `find_auxiliary()` no longer returns `"sigma"` for *brms* models that have no
+  residual standard deviation, but auxiliary parameters whose names merely
+  contain `"sigma"` (e.g. `"sigmadrift"` or `"sigmabias"` from custom families).
+
+* `clean_parameters()` no longer assigns auxiliary parameters whose names
+  contain `"sigma"` (like `"sigmabias"`) to the `"sigma"` component. These are
+  now returned as their own component, which also fixes the related grouping in
+  `parameters::model_parameters()`.
+
 # insight 1.5.4
 
 ## Bug fixes
--- a/R/clean_parameters.R
+++ b/R/clean_parameters.R
@@ -411,24 +411,21 @@ clean_parameters.mlm <- function(x, ...) {
       "fixed"
     }
 
-    # we must start with "conditional one", so it's not confused with "conditional"
-    com <- if (grepl("conditional", i, fixed = TRUE) || i == "random") {
-      "conditional"
-    } else if (grepl("sigma", i, fixed = TRUE)) {
-      "sigma"
-    } else if (grepl("priors", i, fixed = TRUE)) {
-      "priors"
-    } else if (i %in% c("car", "sdcar")) {
-      "car"
-    } else if (grepl("smooth_terms", i, fixed = TRUE)) {
-      "smooth_terms"
-    } else if (grepl("dispersion", i, fixed = TRUE)) {
-      "dispersion"
-    } else if (endsWith(i, "_random")) {
-      gsub("_random$", "", i)
-    } else {
-      i
-    }
+    # elements that hold the group-level part of a component have a "_random"
+    # suffix, e.g. "sigma_random" belongs to the component "sigma" - we strip
+    # that suffix, and then compare the resulting name against the known
+    # components. Note that we must check for *exact* matches here, else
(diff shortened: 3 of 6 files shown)


strengejacke (approved):

strengejacke:
@DominiqueMakowski You can import files from URL with `datawizard::data_read()`, too, and if it's no data frame, it's returned "as is":

``` r
m <- datawizard::data_read(
  "https://raw.github.com/DominiqueMakowski/cogmod/main/vignettes/models/m_ddm.rds"
)
#> Reading data...
#> Imported file is a regression model object of class "brmsfit".
#>   Returning file as is.
```

<sup>Created on 2026-09-08 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

